PR #1, opened 2018-11-30
Removing unncessary example

vanessagraber:


mattpitkin:
I'm not sure how we deal with PRs here on this mirror as development mainly happens on the gitlab page [here](https://git.ligo.org/lscsoft/bilby) (although unfortunately that currently can't take issues/PRs from non-LVC members). @GregoryAshton if this PR looks necessary we should see how we can include it - probably just one of us opening a gitlab MR using the patch from this one.

mattpitkin:
@vanessagraber does having this example in the repo cause a problem, or does the example just seen to be superfluous?

PR #8, opened 2021-11-11
Ensure module name is a str. Not always the case!

AlecThomson:
Just a tiny fix to the log module. In some cases, the module name isn't a string type. This causes an error:
```
TypeError: argument of type 'module' is not iterable
```
This is easily fixed with just a little `str` wrap with no loss in functionality.


GregoryAshton:
Hi @AlecThomson, this has been addressed in https://git.ligo.org/lscsoft/bilby/-/merge_requests/1052, which is v1.1.5, which version of the software are you using?

PR #12, opened 2022-12-16
Made example Tidal Parameters more realistic.

daniellmarc:
Changed the injection values to SLY EoS. Reduced the ranges of delta_lambda, enforced positive components lambdas.

The default settings use unrealistic lambda values that gave problems when checking universal relations. Also realized that the delta lambda range gave a lot of unphysical values so we reduced it to restrict the parameter space. Finally, we propose constraints in the component lambdas so they can be positive, to avoid the issues of negative values in alternative waveform models.


daniellmarc:
Realized we are on the wrong repo. Apologies!

PR #25, opened 2024-02-23
[DEP][DOC}: 696 fix random seed documentation

bretsnev:
Changes:

1. Replaced all instances of `np.random.seed(int)` with `seed(int)`
2. Where relevant added: `from bilby.core.utils.random import seed, rng`, with no rng if its not needed.
3. Added `seed(int)` to "examples/gw_examples/injection_examples/create_your_own_source_model.py", as it uses `rng`, but no seed was set as in other examples.
4. Replaced documentation that asked for legacy numpy.random.state, instead asking for numpy random generator instance and a note about built in `bilby.core.utils.random.rng`

notes:
1. Other instances of `np.random...`  are still in use in the tests instead of using new generator instance.


GregoryAshton:
Thanks @bretsnev this looks great. I reviewed over and I don't see any issues. 

@ColmTalbot should we set the target branch to `github-main`?

examples/gw_examples/injection_examples/change_sampled_parameters.py
```diff
@@ -9,6 +9,10 @@
 """
 import bilby
 import numpy as np
+from bilby.core.utils.random import seed, rng
```
ColmTalbot, line 12: Is this `rng` import needed (also in other scripts where this was added)?

examples/tutorials/compare_samplers.ipynb
```diff
@@ -256,7 +260,25 @@
    "source": []
   }
  ],
- "metadata": {},
+ "metadata": {
```
ColmTalbot, line 263: The metadata should be left empty in jupyter notebooks as it can interfere with others using the notebook, and the automated running in the CI (this is also enforced by the pre-commit checks.)

ColmTalbot (changes requested):
Hi @bretsnev thanks for going through these! I left in a few general comments for minor fixes.

Another thing that I didn't explicitly mention is that in many cases lines have been added/removed, this is another thing that I suspect will be automatically flagged by running the pre-commit checks as `black` formatting is applied to all of the examples.

bretsnev:
Hi @ColmTalbot,

I'm not sure how to run the pre-commit checks here, bit confused with that on github vs looking at the ligo-git. Other than that I've made the other changes you mentioned, thank you for spotting the mistakes :) Please let me know ifyou notice anything else I've missed.

bilby/core/sampler/dynesty_utils.py
```diff
@@ -601,8 +601,9 @@ def propose_differetial_evolution(
         The number of dimensions to run the differential evolution over, the
         first :code:`n_cluster` dimensions are used. The rest are randomly
         sampled from the prior.
-    rstate: np.random.RandomState
-        The random state to use to generate random numbers.
+    rstate: numpy.random.Generator
+        The numpy generator instance used for random number generation.
+		Consider using built in `bilby.core.utils.random.rng`.
```
ColmTalbot, line 606: The CI check is flagging this line (and the one below.) You should align the new line here, I think there's also a distinction between tabs and spaces (we use spaces).

```suggestion
        Consider using built in `bilby.core.utils.random.rng`.
```

examples/core_examples/alternative_samplers/linear_regression_bilby_mcmc.py
```diff
@@ -8,6 +8,10 @@
 import bilby
 import matplotlib.pyplot as plt
 import numpy as np
+from bilby.core.utils.random import seed, rng
```
ColmTalbot, line 11: This is another thing the style checks are failing, imports should be alphabetical.

GregoryAshton:
@ColmTalbot: I can sit with @bretsnev and get the pre-commits set up to fix these issues.

bretsnev:
Thank you for all the help, I think the requested changes have all been made. If you find any more mistakes or if I've missed/made a mistake with any of the previously requested changes, please let me know.

PR #35, opened 2024-09-17
CI: add release workflow

mj-will:
Add a workflow for publishing to PyPI.

This uses new trusted publishers' authentication method.


ColmTalbot (approved):
LGTM!

PR #37, opened 2024-09-17
CI: rename test workflow

ColmTalbot:


mj-will:
I suggest we add the following so CI for previous commits is killed:

```
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true

``

mj-will (approved):
LGTM!

PR #38, opened 2024-09-17
BLD: build docker images

ColmTalbot:


mj-will (approved):

.github/workflows/build-containers.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Build and push Docker image
+        id: push
+        uses: docker/build-push-action@f2a1d5e99d037542a71f64918e516c093c6f3fc4
+        with:
+          context: .
+          tags: ${{ steps.meta.outputs.tags }}
+          push: true
+          labels: ${{ steps.meta.outputs.labels }}
+
+      # - name: Generate artifact attestation
```
adivijaykumar, line 53: I am not super familiar with github workflows, but why are these lines commented? If we don't need them shouldn't we remove them entirely?

ColmTalbot: The short version is that I couldn't get it to work. I think this is a security check that in principle we should probably do. I would probably vote to just add a FIXME comment about adding it back in later.

GregoryAshton (approved):

.github/workflows/build-containers.yml
adivijaykumar: OK, sounds good. I would be in favour of adding a FIXME comment.

ColmTalbot:
```suggestion
      # FIXME: disabled as it is not working well but should probably be added in later
      # https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds
      # - name: Generate artifact attestation
```

adivijaykumar (approved):

Issue #41, opened 2024-09-23
Error in warning issued when enforcing LAL kwargs 


jacobgolomb:
The warning introduced by this MR: https://git.ligo.org/lscsoft/bilby/-/merge_requests/1269 is printed by this line https://git.ligo.org/lscsoft/bilby/-/blob/master/bilby/gw/source.py?ref_type=heads#L672.

This raises a KeyError because I don't think you can print a dict with the `.format` method like this. I think it needs to be changed to `.format(waveform_kwargs=waveform_kwargs)`.

ColmTalbot:
Closing as resolved

PR #42, opened 2024-09-23
BUG: changed formatting for unused kwargs error message

jacobgolomb:
fixes issue #41

### DIFF
--- a/test/gw/source_test.py
+++ b/test/gw/source_test.py
@@ -78,7 +78,7 @@ class TestLalBBH(unittest.TestCase):
 
         with self._caplog.at_level(logging.WARNING, logger="bilby"):
             bilby.gw.source.lal_binary_black_hole(
-                    self.frequency_array, **self.parameters
+                self.frequency_array, **self.parameters
             )
             assert "There are unused waveform kwargs" in self._caplog.text
 
--- a/test/gw/source_test.py
+++ b/test/gw/source_test.py
@@ -1,4 +1,6 @@
 import unittest
+import logging
+import pytest
 
 import bilby
 import lal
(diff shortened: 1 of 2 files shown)


jacobgolomb:
> It looks like there is support in pytest to capture log messages, e.g., [here](https://docs.pytest.org/en/7.1.x/how-to/logging.html).

Instead I just captured the logger output with the unittest context manager and added an assertion that the expected warning message is in the logger's output. Is that an okay alternative?

ColmTalbot (approved):
The new test looks good to me, happy to merge once the CI passes.

jacobgolomb:
> The new test looks good to me, happy to merge once the CI passes.

The unit test is failing in the CI, but passes locally for me:
```
>>> from source_test import TestLalBBH
>>> test = TestLalBBH()
>>> test.setUp()
>>> test.test_unused_waveform_kwargs_message()
```

jacobgolomb:
> > The new test looks good to me, happy to merge once the CI passes.
> 
> The unit test is failing in the CI, but passes locally for me:
> 
> ```
> >>> from source_test import TestLalBBH
> >>> test = TestLalBBH()
> >>> test.setUp()
> >>> test.test_unused_waveform_kwargs_message()
> ```

Maybe this could be due to self.parameters being updated before this specific test is run in CI and causing a different error to be raised first.

jacobgolomb:
I thought I fixed it; running `pytest source_test.py::TestLalBBH` works locally, with the same versions of pytest.

mj-will:
Sorry, I wanted to test a few things and accidentally push to this branch instead of my own. I've reset the changes.

test/gw/source_test.py
```diff
@@ -65,6 +71,21 @@ def test_waveform_error_raising(self):
                 self.frequency_array, **raise_error_parameters
             )
 
+    def test_unused_waveform_kwargs_message(self):
+        self.parameters.update(self.waveform_kwargs)
+        self.parameters["unused_waveform_parameter"] = 1.0
+        self._caplog.set_level(logging.WARNING, logger="bilby")
+        bilby.gw.source.logger.setLevel(logging.WARNING)
```
mj-will, line 78: I don't think these lines are necessary, since you set them below in the `with` statement.

```suggestion

```

jacobgolomb:
Thanks @mj-will for figuring out the issue with the test! I'll rebase after #855 is merged so that this passes the CI.

mj-will (approved):

Issue #31, opened 2024-08-01
Subclasses of JointPrior are re-initialized wrong from dictionaries and json


JasperMartins:
I am not a member of the LVK and, thus, cannot create issues or merge requests in the main repository.

I have encountered issues with subclassing JointPrior and JointPriorDist that occur when initializing from `dict` and `json` representations, as done when loading a bilby results object. In both instances, checks for `MultivariateGaussin(Dist)` are hardcoded without need, as far as I understand it. Changing these checks to `(Base)JointPrior(Dist)` solves the problems.

For `json`, this block

https://github.com/lscsoft/bilby/blob/bad30b7cfd732d16b62322e45db5bb5f9ebc4f3c/bilby/core/utils/io.py#L29-L47

could be updated to
```python
class BilbyJsonEncoder(json.JSONEncoder):
    def default(self, obj):
        from ..prior import BaseJointPriorDist, Prior, PriorDict
        from ...gw.prior import HealPixMapPriorDist
        from ...bilby_mcmc.proposals import ProposalCycle

        if isinstance(obj, np.integer):
            return int(obj)
        if isinstance(obj, np.floating):
            return float(obj)
        if isinstance(obj, PriorDict):
            return {"__prior_dict__": True, "content": obj._get_json_dict()}
        if isinstance(obj, (BaseJointPriorDist, HealPixMapPriorDist, Prior)):
``` 

For `dict`s, 

https://github.com/lscsoft/bilby/blob/bad30b7cfd732d16b62322e45db5bb5f9ebc4f3c/bilby/core/prior/dict.py#L208-L308

could be updated to
```python
def from_dictionary(self, dictionary):
        jpdkwargs = {}
        for key in list(dictionary.keys()):
            val = dictionary[key]
            if isinstance(val, Prior):
                continue
            elif isinstance(val, (int, float)):
                dictionary[key] = DeltaFunction(peak=val)
            elif isinstance(val, str):
                cls = val.split("(")[0]
                args = "(".join(val.split("(")[1:])[:-1]
                try:
                    dictionary[key] = DeltaFunction(peak=float(cls))
                    logger.debug("{} converted to DeltaFunction prior".format(key))
                    continue
                except ValueError:
                    pass
                if "." in cls:
                    module = ".".join(cls.split(".")[:-1])
                    cls = cls.split(".")[-1]
                else:
                    module = __name__.replace(
                        "." + os.path.basename(__file__).replace(".py", ""), ""
                    )
                try:
[...]

PR #44, opened 2024-09-26
BUGFIX: Joint prior subclassing fixes

JasperMartins:
This pull request addresses multiple issues with the subclassing of  BaseJointPrior lists and their reinitialization from `json` or `dict` outlined in https://github.com/bilby-dev/bilby/issues/31, implements fixes and also test cases that fail with the old implementation.

The tests uncovered some new issues that required fixes beyond the (current) discussion in https://github.com/bilby-dev/bilby/issues/31. The fixes are implemented by moving some of the logic used to re-initialize (subclasses of) `BaseJointPriorDist` from `MultivariateGaussianDist` to `BaseJointPriorDist`.

The tests also uncovered a problem with `__eq__` for `JointPrior` that resulted in `__eq__` giving `False` if the `dist` keyword does not point to the same instance. This, I felt, was inconsistent with the usage of `__eq__` across `PriorDict`. 

One overall change currently applied by this PR is that it is now easier to, by mistake, use a not-subclassed `BaseJointPriorDist` somewhere. I wonder, however, if the `BaseJointPriorDist` methods that need be subclassed should not raise an `NotImplemented` error anyway.

### DIFF
--- a/bilby/core/prior/dict.py
+++ b/bilby/core/prior/dict.py
@@ -8,7 +8,7 @@ import numpy as np
 
 from .analytical import DeltaFunction
 from .base import Prior, Constraint
-from .joint import JointPrior
+from .joint import JointPrior, BaseJointPriorDist
 from ..utils import (
     logger,
     check_directory_exists_and_if_not_mkdir,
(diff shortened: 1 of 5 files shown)


ColmTalbot (approved):
Thanks @JasperMartins I'm happy with these changes and it looks like they are well covered in tests.

On the point about being able to instantiate the base distribution. We generally haven't been good about making base classes unusable, it would be nice longer term to track down the various places we should either have some kind of abstract base or a bunch of `NotImplementedError`s.

mattpitkin (approved):
Looks good to me

PR #47, opened 2024-10-02
FEAT: add argument to specify calibration correction direction

ColmTalbot:
This PR aims to add sufficient flexibility in reading calibration data products to make sure that we correctly map these to the templates in the likelihood. There are probably additional documentation changes that we will want.

### DIFF
--- a/bilby/gw/detector/calibration.py
+++ b/bilby/gw/detector/calibration.py
@@ -12,7 +12,7 @@ from ...core.prior.dict import PriorDict
 from ..prior import CalibrationPriorDict
 
 
-def read_calibration_file(filename, frequency_array, number_of_response_curves, starting_index=0):
+def read_calibration_file(filename, frequency_array, number_of_response_curves, starting_index=0, correction=None):
     r"""
     Function to read the hdf5 files from the calibration group containing the
     physical calibration response curves.
@@ -33,6 +33,9 @@ def read_calibration_file(filename, frequency_array, number_of_response_curves, 
     starting_index: int
         Index of the first curve to use within the array. This allows for segmenting the calibration curve array
         into smaller pieces.
+    correction: str
+        How the correction is defined, either to the data (default) or
+        the template, see :func:`bilby.gw.prior.CalibrationPriorDict.from_envelope_file`.
 
     Returns
     -------
@@ -42,6 +45,21 @@ def read_calibration_file(filename, frequency_array, number_of_response_curves, 
 
     """
     import tables
+
+    if correction is None:
+        logger.warning(
+            "Calibration envelope correction type is not specified. "
+            "Assuming this correction maps calibrated data to theoretical "
+            "strain. If this is correct, this should be explicitly "
+            "specified via CalibrationPriorDict.from_envelope_file(..., "
+            "correction='data')."
+        )
+        correction = "data"
+    if correction.lower() not in ["data", "template"]:
+        raise ValueError(
+            "Calibration envelope correction should be one of 'data' or "
+            f"'template', found {correction}."
+        )
 
     logger.info(f"Reading calibration draws from {filename}")
     calibration_file = tables.open_file(filename, 'r')
@@ -61,9 +79,11 @@ def read_calibration_file(filename, frequency_array, number_of_response_curves, 
 
     # interpolate to the frequency array (where if outside the range of the calibration uncertainty its fixed to 1)
     calibration_draws = calibration_amplitude * np.exp(1j * calibration_phase)
-    calibration_draws = 1 / interp1d(
+    calibration_draws = interp1d(
         calibration_frequencies, calibration_draws, kind='cubic',
         bounds_error=False, fill_value=1)(frequency_array)
+    if correction == "data":
+        calibration_draws **= -1
 
     try:
         parameter_draws = pd.read_hdf(filename, key="CalParams")
@@ -73,7 +93,7 @@ def read_calibration_file(filename, frequency_array, number_of_response_curves, 
     return calibration_draws, parameter_draws
 
 
-def write_calibration_file(filename, frequency_array, calibration_draws, calibration_parameter_draws=None):
+def write_calibration_file(filename, frequency_array, calibration_draws, calibration_parameter_draws=None, correction=None):
     """
     Function to write the generated response curves to file
 
@@ -88,9 +108,32 @@ def write_calibration_file(filename, frequency_array, calibration_draws, calibra
         Shape is (number_of_response_curves x len(frequency_array))
     calibration_parameter_draws: data_frame
         Parameters used to generate the random draws of the calibration response curves
+    correction: str
+        How the correction is defined, either to the data (default) or
+        the template, see :func:`bilby.gw.prior.CalibrationPriorDict.from_envelope_file`.
+        If :code:`data`, the calibration draws will be inverted before being saved so that the format
+        matches files produced by detector calibration pipelines.
 
     """
     import tables
+
+    if correction is None:
+        logger.warning(
+            "Calibration envelope correction type is not specified. "
+            "Assuming this correction maps calibrated data to theoretical "
+            "strain. If this is correct, this should be explicitly "
+            "specified via CalibrationPriorDict.from_envelope_file(..., "
+            "correction='data')."
+        )
+        correction = "data"
+    if correction.lower() not in ["data", "template"]:
+        raise ValueError(
+            "Calibration envelope correction should be one of 'data' or "
+            f"'template', found {correction}."
+        )
+
+    if correction == "data":
+        calibration_draws **= -1
 
     logger.info(f"Writing calibration draws to {filename}")
     calibration_file = tables.open_file(filename, 'w')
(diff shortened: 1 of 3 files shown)


mj-will:
We had some discussion about this on the last dev call but not many of us attended. So I'd encourage people to comment.

@asb5468 could you add a comment with what you said on the call?

farr:
@ColmTalbot FWIW, I like this change, particularly the description and choices in the keyword argument `correction='data'` or `correction='template'`.  One thing I noticed which is subtle (but, as written, correct!) is in the reading of the calibration uncertainties in phase and amplitude: currently the code *assumes* the "ordering" of the quantiles given in the calibration file and performs subtraction to determine the `sigma` parameters in the calibration envelope (ordering reversed for `data` calibration where you are taking the inverse):
```python
        if correction.lower() == "data":
            amplitude_median = 1 / calibration_data[1] - 1
            phase_median = -calibration_data[2]
            amplitude_sigma = (1 / calibration_data[3] - 1 / calibration_data[5]) / 2
            phase_sigma = (calibration_data[6] - calibration_data[4]) / 2
        else:
            amplitude_median = calibration_data[1] - 1
            phase_median = calibration_data[2]
            amplitude_sigma = (calibration_data[5] - calibration_data[3]) / 2
            phase_sigma = (calibration_data[6] - calibration_data[4]) / 2
```
I might throw an `np.abs(...)` around those, just to be robust against changing the ordering (ultimately, it would be nice to have the columns named, instead of indexed, so you can just do `(calibration_data['amplitude_q84'] - calibration_data['amplitude_q16'])/2` and don't have to rely on opaque column numbers).

transientlunatic:
This lgtm.

bilby/gw/detector/calibration.py
```diff
@@ -12,9 +12,15 @@
 from ..prior import CalibrationPriorDict
 
 
-def read_calibration_file(filename, frequency_array, number_of_response_curves, starting_index=0):
-    """
-    Function to read the hdf5 files from the calibration group containing the physical calibration response curves.
+def read_calibration_file(filename, frequency_array, number_of_response_curves, starting_index=0, correction=None):
```
asb5468, line 15: This is a minor point, but can the new parameter be called `correction_type` rather than `correction`? The latter to me seems like an optional boolean variable, like whether to implement a higher-order correction.

bilby/gw/detector/calibration.py
```diff
@@ -27,6 +33,9 @@ def read_calibration_file(filename, frequency_array, number_of_response_curves,
     starting_index: int
         Index of the first curve to use within the array. This allows for segmenting the calibration curve array
         into smaller pieces.
+    correction: str
+        How the correction is defined, either to the data (default) or
```
cplb, line 37: I'd suggest either (i) we make `correction='data'` instead of `None`, or (ii) change the text here to say something like `traditional` or `usual` rather than `default`. I suspect (ii) is better, as we want to avoid cases where people don't know which one they are doing.

bilby/gw/detector/calibration.py
```diff
@@ -82,10 +110,33 @@ def write_calibration_file(filename, frequency_array, calibration_draws, calibra
         Shape is (number_of_response_curves x len(frequency_array))
     calibration_parameter_draws: data_frame
         Parameters used to generate the random draws of the calibration response curves
+    correction: str
+        How the correction is defined, either to the data (default) or
```
cplb, line 114: Same comment here as on L37 https://github.com/bilby-dev/bilby/pull/47/files#r1795826808

bilby/gw/detector/calibration.py
```diff
@@ -12,9 +12,15 @@
...
-def read_calibration_file(filename, frequency_array, number_of_response_curves, starting_index=0):
-    """
-    Function to read the hdf5 files from the calibration group containing the physical calibration response curves.
+def read_calibration_file(filename, frequency_array, number_of_response_curves, starting_index=0, correction=None):
+    r"""
+    Function to read the hdf5 files from the calibration group containing the
+    physical calibration response curves.
+
+    These curves are defined to convert calibrated strain (:code:`d`) to
+    theoretical waveform templates (:code:`h`), in the likelihood we apply
```
adivijaykumar, line 21: Shouldn't this be theoretical strain and not theoretical waveform template?

bilby/gw/detector/calibration.py
```diff
@@ -37,6 +46,21 @@ def read_calibration_file(filename, frequency_array, number_of_response_curves,
     """
     import tables
 
+    if correction is None:
+        logger.warning(
+            "Calibration envelope correction type is not specified. "
+            "Assuming this correction maps calibrated data to theoretical "
+            "strain. If this is correct, this should be explicitly "
+            "specified via CalibrationPriorDict.from_envelope_file(..., "
+            "correction='data')."
```
adivijaykumar, line 55: Perhaps we should also discuss what to do when `correction='data'` is not the intended operation?

bilby/gw/detector/calibration.py
```diff
@@ -82,10 +110,33 @@ def write_calibration_file(filename, frequency_array, calibration_draws, calibra
...
+    correction: str
+        How the correction is defined, either to the data (default) or
+        the template, see :func:`bilby.gw.prior.CalibrationPriorDict.from_envelope_file`.
+        If :code:`data`, the calibration draws will be inverted before being saved so that the format
+        matches files produced by detector calibration pipelines.
 
     """
     import tables
 
+    if correction is None:
```
adivijaykumar, line 122: There is some code duplication here. These statements are used three times in total. Would it be possible to wrap this to a function like `_check_correction_type(correction)` or something similar?

adivijaykumar (commented):
Broadly LGTM! I have left a few comments; I also agree with suggestions raised by Sylvia and Will.

bilby/gw/prior.py
```diff
@@ -1138,7 +1138,7 @@ def to_file(self, outdir, label):
     @staticmethod
     def from_envelope_file(envelope_file, minimum_frequency,
                            maximum_frequency, n_nodes, label,
-                           boundary="reflective"):
+                           boundary="reflective", correction=None):
```
jacobgolomb, line 1141: Agree with `correction` -> `correction_type`, but also I think the default in the signature should be `correction_type='data'` and there should just be a message

mj-will:
There are a few comments about whether the default should be `None` or `'data'`. FWIW, I'm in favour of keeping it as `None` for now with a warning, as it makes it very clear that you _should_ be setting it.

In future, I'd be in favour of going a step further and either making it a keyword argument *without* a default, meaning the function cannot be called without specifying the type of correction being used. This would break any codes which do not set it and avoid potential issues due to assumed behaviour/ignoring warnings.

bilby/gw/detector/calibration.py
ColmTalbot: Is the issue here that `None` is mapped internally to `data` ad so `data` technically isn't the default value? I'm not really concerned about that as practically the default is `data`.

bilby/gw/detector/calibration.py
ColmTalbot: I actually removed this whole passage as I added the `correction_type` argument. Do you think some amount of this should be added to the docstring?

bilby/gw/detector/calibration.py
```diff
@@ -27,6 +91,13 @@ def read_calibration_file(filename, frequency_array, number_of_response_curves,
     starting_index: int
         Index of the first curve to use within the array. This allows for segmenting the calibration curve array
         into smaller pieces.
+    correction_type: str
+        How the correction is defined, either to the :code:`data`
+        (default) or the :code:`template`. In general, data products
+        produced by the LVK calibration groups assume :code:`data`.
+        The default value will be removed in a future release and
+        this will need to be explicitly specified.
+        ..versionadded XYZ
```
adivijaykumar, line 100: You were probably already planning on changing this, but reminder to change this to the actual version.

adivijaykumar (approved):

bilby/gw/detector/calibration.py
adivijaykumar: I like the latest version!

mj-will (approved):

bilby/gw/prior.py
```diff
@@ -1149,6 +1149,14 @@ def from_envelope_file(envelope_file, minimum_frequency,
             frequency median-amplitude median-phase -1-sigma-amplitude
             -1-sigma-phase +1-sigma-amplitude +1-sigma-phase
 
+        There are two definitions of the calibration correction in the
+        literature, one defines the correction as mapping calibrated strain
+        to theoretical waveform templates (:code:`data`) and the other as
+        mapping theoretical waveform templates to calibrated strain
+        (:code:`template`). Prior to version XYZ, :code:`template` was assumed,
```
adivijaykumar, line 1156: This line is missing the version number. Once that is fixed I am happy to approve.

```suggestion
        (:code:`template`). Prior to version 1.4.0, :code:`template` was assumed,
```

adivijaykumar (approved):

PR #823, opened 2024-10-07
BUG: Fix arviz bug import bug

mj-will:
Closes https://github.com/bilby-dev/bilby/issues/799

Original GitLab MR: https://git.ligo.org/lscsoft/bilby/-/merge_requests/1382

### DIFF
--- a/bilby/core/result.py
+++ b/bilby/core/result.py
@@ -1707,13 +1707,17 @@ class Result(object):
             =======
             azdata: InferenceData
                 The ArviZ InferenceData object.
+
+            Raises
+            ======
+            RuntimeError: If ArviZ is not installed.
         """
 
         try:
             import arviz as az
         except ImportError:
-            logger.debug(
-                "ArviZ is not installed, so cannot convert to InferenceData"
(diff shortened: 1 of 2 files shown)


adivijaykumar (approved):

bilby/core/result.py
```diff
@@ -1707,13 +1707,17 @@ def to_arviz(self, prior=None):
...
+            ======
+            RuntimeError: If ArviZ is not installed.
         """
 
         try:
             import arviz as az
         except ImportError:
-            logger.debug(
-                "ArviZ is not installed, so cannot convert to InferenceData"
+            raise RuntimeError(
```
ColmTalbot, line 1719: I'm not sure `RuntimeError` is the best choice here. I see this as kind of a last-ditch kind of error when nothing else makes sense. We have a [`ResultError`](https://github.com/bilby-dev/bilby/blob/main/bilby/core/result.py#L2238). Or we could just raise the current debug to a `logger.error` and raise the original exception.

ColmTalbot (approved):

PR #827, opened 2024-10-08
Add Neural Likelihood Estimation (NLE) likelihood

GregoryAshton:
This PR adds a method to utilise NLE within bilby

Summary of changes:
1. Refactor the `core/likelihood.py` module into a set of submodules
2. Add a module core/likelihood/simulation_based_inference.py


mj-will:
Hi @GregoryAshton, I haven't had a change to look over this properly but it seems like a pretty bit addition. Given that, I thought it might be worth discussing whether this should go into `bilby` or perhaps a downstream package?

My main concern is the additional maintenance burden from having code that depends on `sbi`. This seems somewhat opposite to what we've been doing with the samplers. What do you think?

Happy to discuss this on a call at some point.

GregoryAshton:
@mj-will Yep, understood. This is just a draft at the moment. I wanted to make it public, and I can bring it up on a call to discuss and make others aware, but where it ultimately lives certainly needs some thought.

However, for now I can say that my intention (if it lives in Bilby) is to make it an optional dependency (which is why I refactored the likelihood to enable this). Right now I am mainly working on the implementation.

mj-will:
@GregoryAshton sounds good. I've marked this as a draft for now, so it's clear you're still working on this.

mj-will:
@MattiaEmma2022 @GregoryAshton should this be closed now that [sbilby](https://github.com/MattiaEmma2022/sbilby/) exists?

Issue #830, opened 2024-10-11
GW190425 example fails due to a relative binning error


mj-will:
The [GW190425 example](https://github.com/bilby-dev/bilby/blob/main/examples/gw_examples/data_examples/GW190425.py) currently fails with the following error that is related to relative binning:

```
Traceback (most recent call last):
  File "/home/mjwill/git_repos/bilby/examples/gw_examples/data_examples/GW190425.py", line 118, in <module>
    likelihood = bilby.gw.likelihood.RelativeBinningGravitationalWaveTransient(
  File "/home/mjwill/git_repos/bilby/bilby/gw/likelihood/relative.py", line 148, in __init__
    self.compute_summary_data()
  File "/home/mjwill/git_repos/bilby/bilby/gw/likelihood/relative.py", line 334, in compute_summary_data
    stop = masked_frequency_array[end_idx]
IndexError: index 62977 is out of bounds for axis 0 with size 62977
```

I suspect this means there's a bug in the relative binning likelihood.

mj-will:
I've checked, and this example still ran in v2.2.2, so I believe this bug was introduced in https://github.com/bilby-dev/bilby/commit/103a5127b3f896e924fd77e3bbcbbc9b404fb9ca#diff-b8983bc349cc3e1bc4986b6de834bf9493ef07528d022f86510032dce710718bR219.

@adivijaykumar do you have any insight into this?

ColmTalbot:
Is this addressed by #48?

mj-will:
> Is this addressed by #48?

Ah, I'd missed that. I think it is.

mj-will:
Closing this since #48 has been merged.

PR #48, opened 2024-10-02
[BUG] relative binning - correctly set last bin index

noahewolfe:
If the last bin index is already chosen to be the last point in the frequency array, we shouldn't increment by 1, else we have an invalid index.

I observed this with very long injections that have power even at the maximum frequency, and thus, the bin selection will already include the last frequency point in the bin indices.

PR #831, opened 2024-10-11
BUG: fix relative binning indices

mj-will:
In some cases, the relative binning likelihood would fail due the final index being invalid.

I think this is all that is needed to fix it but others should comment. I've tested the GW190425 example, and it runs after this change.

Closes https://github.com/bilby-dev/bilby/issues/830


mj-will:
This is a duplicated of https://github.com/bilby-dev/bilby/pull/48

Issue #836, opened 2024-10-23
Bug: bilby_pipe default latex labels cause bilby plotting error downstream


AlexandreGoettel:
When using bilby_pipe with very generic settings and default labels, the "\" symbols in those get duplicated at some point, meaning that in the result.hdf5 file, you will get, for example, "$\\mathcal{M}$" instead of"$\mathcal{M}$". I am guessing it is due to conversion but am unsure when that happens.

I suspect (I am more or less guessing.. correct me if I'm wrong!) that some of the recent matplotlib changes are now causing this to throw errors. I am using matplotlib 3.7.3.
I am opening this issue on the bilby repository because I noticed that there is already a function to modify latex labels to remove matplotlib errors, so I figured it would be a good idea to simply expand this function.

<details open>
  <summary>Full error Traceback</summary>

```
Traceback (most recent call last):
  File "/home/gottela/Software/miniforge3/envs/sbi/bin/bilby_pipe_plot_marginal", line 10, in <module>
    sys.exit(plot_marginal())
             ^^^^^^^^^^^^^^^
  File "/home/gottela/Software/miniforge3/envs/sbi/lib/python3.11/site-packages/bilby_pipe/plot.py", line 133, in plot_marginal
    result.plot_marginals(priors=True)
  File "/home/gottela/Software/bilby/bilby/core/result.py", line 1138, in plot_marginals
    self.plot_single_density(
  File "/home/gottela/Software/bilby/bilby/core/utils/plotting.py", line 65, in wrapper_decorator
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/gottela/Software/bilby/bilby/core/result.py", line 1054, in plot_single_density
    fig.tight_layout()
  File "/home/gottela/Software/miniforge3/envs/sbi/lib/python3.11/site-packages/matplotlib/figure.py", line 3543, in tight_layout
    engine.execute(self)
  File "/home/gottela/Software/miniforge3/envs/sbi/lib/python3.11/site-packages/matplotlib/layout_engine.py", line 184, in execute
    kwargs = get_tight_layout_figure(
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gottela/Software/miniforge3/envs/sbi/lib/python3.11/site-packages/matplotlib/_tight_layout.py", line 266, in get_tight_layout_figure
    kwargs = _auto_adjust_subplotpars(fig, renderer,
[...]

PR #837, opened 2024-10-23
Remove double-backslashes from latex labels passed to matplotlib.

AlexandreGoettel:
Addresses #836 
As far as I can tell, the issue is due to a bug in bilby_pipe, but I noticed that there is already a function to remove matplotlib errors in bilby/core/result.py.

This PR simply adds a line to this function to convert "\\" into "\".

Either way I think this is a good thing to have, as I can't think of a scenario where having "\\" in a matplotlib latex string is a good idea.

### DIFF
--- a/bilby/core/result.py
+++ b/bilby/core/result.py
@@ -1023,6 +1023,7 @@ class Result(object):
         import matplotlib.pyplot as plt
         logger.info('Plotting {} marginal distribution'.format(key))
         label = self.get_latex_labels_from_parameter_keys([key])[0]
+        label = sanity_check_labels([label])[0]
         fig, ax = plt.subplots()
         try:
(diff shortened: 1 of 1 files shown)


ColmTalbot (approved):

mj-will (commented):
Have you checked to see how this interacts with raw strings (e.g. `r"$\mathcal{M}$"`)?

mj-will:
> Not 100% what you mean but afaik a variable defined as a raw string just becomes a string (with different creation rules) i.e.: `r"$\mathcal{M}$" == "$\\mathcal{M}$" will return True.` The code above will work in both cases.

Ah, I didn't know that, makes sense then :)

mj-will (approved):

PR #839, opened 2024-10-25
Draft: Added qmc_quad based method for estimation of the constrained normalization factor

JasperMartins:
This PR implements a new method to estimate the normalization factor for constrained priors.
The changes are two-fold:

1. The integration is performed with `scipy.integrate.qmd_quad`, a quasi-Monte Carlo-based integration routine that is expected to yield better results than regular Monte Carlo integration. However, the routine requires a rescaling step from the unit cube rather than direct sampling. 
2. The termination of the integration is based on its relative statistical error rather than the number of accepted samples.

I have tested the two implementations with a relatively easy scenario: A 2D uniform prior on the [-1,1] cube, constrained to an inscribed circle with different radii: 

image

The new method is significantly faster for high normalization factors, and the relative errors show a similar spread.

The implementation is marked as a draft because of the requirement of a `rescale`-method of the priors. Thus, it could be nice to keep the old method as a fallback. Also, the relative-error termination criterion could be applied just as well to the old implementation.

### DIFF
--- a/bilby/core/prior/dict.py
+++ b/bilby/core/prior/dict.py
@@ -5,6 +5,7 @@ from importlib import import_module
 from io import open as ioopen
 
 import numpy as np
+from scipy.integrate import qmc_quad
 
 from .analytical import DeltaFunction
 from .base import Prior, Constraint
@@ -464,6 +465,44 @@ class PriorDict(dict):
                 key: np.reshape(all_samples[key][:needed], size) for key in keys
             }
             return all_samples
+
+    def normalize_constraint_factor_qmc_quad(self, keys, nrepeats=10, sampling_chunk=50000,
+                                             rel_error_target=0.01, max_trials=False):
+
+        def integrand(theta):
+            samples = np.array(self.rescale(keys=keys, theta=theta)).reshape((len(keys), -1))
+            samples = {key: samps for key, samps in zip(keys, samples)}
+            probs = self.evaluate_constraints(samples)
+            return probs
+
+        trials = 0
+        estimates = []
+        weights = []
+        while True:
+            res = qmc_quad(func=integrand, a=np.zeros(len(keys)), b=np.ones(len(keys)),
+                           n_estimates=nrepeats, n_points=sampling_chunk)
+            trials += sampling_chunk * nrepeats
+
+            # compute a weighted mean of the integral measurements and the resulting standard deviation
+            estimates.append(res.integral)
+            weights.append(1 / res.standard_error)
+
+            cumulative_weighted_error = 1 / np.sum(weights)
+            cumulative_weighted_integral = cumulative_weighted_error * np.sum(np.array(estimates) * np.array(weights))
+
+            # compute the rounded factor and relative error (as given by the standard deviation)
+            factor = 1 / cumulative_weighted_integral
+            rel_error = factor * cumulative_weighted_error
+
+            if rel_error > rel_error_target:
+                if max_trials and (trials + sampling_chunk * nrepeats > max_trials):
+                    break
+            else:
+                break
+
+        decimals = int(-np.floor(np.log10(3 * rel_error)))
+        factor_rounded = np.round(factor, decimals)
+        return factor_rounded
 
     def normalize_constraint_factor(
         self, keys, min_accept=10000, sampling_chunk=50000, nrepeats=10


JasperMartins:
I have updated the PR quite a bit. The core logic of the integration of the normalization factor is now handled by one of two functions: either MC-Integration based on samples from `PriorDict.sample`, or quasi MC-Integration based on the `rescale` method. The user can choose which is used, but the code also checks if the `rescale` method is implemented if `qmc_quad` is used and will default to `from_samples` if not. For both methods, termination of the integration is handled vi a bound of the estimated relative error. For both methods, a `max_trials` kwarg can be used to limit the number of probability evaluations.

I have also optimized the `from_samples` implementation. Before, every time `min_accept` was not reached, new samples were added to a list, and the constrained was applied to the *full* list - yielding a steep increase in runtime with the number of iterations while it would have been sufficient to check the new samples.

For the example I gave above, the qmc-based implementation is now actually slower than the `from_samples` method due to a higher overhead. Priors that implement `sample` by just calling `rescale` on unit-samples should perform much closer.
I still selected `qmc_quad` as the default because, as the attached plot shows, for normalization factors close to 1 (which is more likely in most applications), the relative error is smaller.  

I have also improved the robustness against bugs by checking if the chosen keys are sufficient to compute the constrained, and if the PriorDict is constrained in the first place.

image

bilby/core/prior/dict.py
```diff
@@ -465,39 +466,122 @@ def sample_subset_constrained(self, keys=iter([]), size=None):
...
+
+        if method == "qmc_quad":
+            integrator = self._integrate_normalization_factor_from_qmc_quad
+        elif method == "from_samples":
+            integrator = self._integrate_normalization_factor_from_samples
+        else:
+            raise ValueError(f"Integration method {method} not understood")
+        try:
+            theta = np.random.uniform(0, 1, size=(len(keys), 1))
+            samples = {key: self[key].rescale(units) for key, units in zip(keys, theta)}
```
ColmTalbot, line 549: This will probably fail for conditional priors, the `self.rescale` method should be safer (e.g., [here](https://github.com/bilby-dev/bilby/blob/main/bilby/core/prior/dict.py#L596))

bilby/core/prior/dict.py
```diff
@@ -465,39 +466,122 @@ def sample_subset_constrained(self, keys=iter([]), size=None):
...
+            for key in keys:
+                if np.any(np.isnan(samples[key])):
+                    print("The rescale method appears to be not working. Switching to 'sample_based'.")
+                    integrator = self._integrate_normalization_factor_from_samples
+        except Exception:
```
ColmTalbot, line 554: I'd rather not use a plain `Exception`. Do you have an example of a case that failed and what error is raised? I would imagine it would be something like `NotImplementedError` or `AttributeError`?

bilby/core/prior/dict.py
```diff
@@ -519,6 +603,9 @@ def prob(self, sample, **kwargs):
         return self.check_prob(sample, prob)
 
     def check_prob(self, sample, prob):
+        if self.conversion_function == self.default_conversion_function:
```
ColmTalbot, line 606: I don't think this is safe, some of the other `PriorDict` subclasses have non-trivial default conversions, e.g., [`BBHPriorDict`](https://github.com/bilby-dev/bilby/blob/main/bilby/gw/prior.py#L889). Is there a specific reason you added this?

bilby/core/prior/dict.py
```diff
@@ -558,10 +645,10 @@ def ln_prob(self, sample, axis=None, normalized=True):
                                   normalized=normalized)
 
     def check_ln_prob(self, sample, ln_prob, normalized=True):
-        if normalized:
+        if normalized and (self.conversion_function != self.default_conversion_function):
```
ColmTalbot, line 648: Again, I don't think this is something we can assume.

bilby/core/prior/dict.py
```diff
@@ -465,39 +466,122 @@ def sample_subset_constrained(self, keys=iter([]), size=None):
...
+        while True:
+            for i in range(nrepeats):
+                integral = integrator(sample_keys, sampling_chunk, **kwargs)
+                trials += sampling_chunk
+
+                # compute a weighted mean of the integral measurements and the resulting standard deviation
+                estimates.append(integral)
+
+            standard_error = np.std(estimates, ddof=1) / np.sqrt(len(estimates))
+            cumulative_integral = np.sum(estimates) / len(estimates)
```
ColmTalbot, line 569: Can this just be?

```suggestion
            cumulative_integral = np.mean(estimates)
```

ColmTalbot (changes requested):
This is looking good! I just have a few specific comments/questions.

bilby/core/prior/dict.py
JasperMartins: I wanted to check if the normalization factor integration is necessary by checking if the prior is constrained because if not the new error estimation will fail. However, this should be handled better on the level of the error estimation anyway because it will also fail at the moment of the constraint does not affect the list of keys selected.

Are there situations where checking if any prior in the PriorDict is a constraint and only applying the normalization if so would be unsafe?

bilby/core/prior/dict.py
JasperMartins: It's probably better completely removed. I thought about user-side errors in the implementation of custom priors, but such cases should probably fail loudly rather than silently.

Maybe catching `NotImplementedError` could be added to prepare for later changes that add `NotImplementedError`s to the base classes, but that would also render the check if some priors yield `None` unnecessary.

ColmTalbot: This sounds good to me.

bilby/core/prior/dict.py
ColmTalbot: > Are there situations where checking if any prior in the PriorDict is a constraint and only applying the normalization if so would be unsafe?

This should be a sufficient check. You can query the length of [`PriorDict.constraint_keys`](https://github.com/bilby-dev/bilby/blob/main/bilby/core/prior/dict.py#L440). (It might also be useful to add a `has_constraints` or similar property.)

bilby/core/prior/dict.py
JasperMartins: I updated the PR so that both cases are accounted for

bilby/core/prior/dict.py
```diff
@@ -465,39 +471,125 @@ def sample_subset_constrained(self, keys=iter([]), size=None):
...
+                theta = np.random.uniform(0, 1, size=(len(keys), 1))
+                samples = self.rescale(keys, theta)
+                none_keys = []
+                for i, key in enumerate(keys):
+                    if samples[i] is None:
+                        none_keys.append(key)
+                if len(none_keys) > 0:
+                    raise NotImplementedError(f"The rescale method returns 'None', for key(s) {none_keys}.")
+            except NotImplementedError as e:
+                print(f"The rescaling step fails with message:\n{e}")
```
ColmTalbot, line 555: These print statements should use `logger.info`.

bilby/core/prior/dict.py
```diff
@@ -465,39 +471,125 @@ def sample_subset_constrained(self, keys=iter([]), size=None):
...
+            most cases. However, 'qmc_quad' is expected to yield lower errors against the ground truth for cases with
+            few repeats, but high sampling_chunk.
+
+        Returns
+        =======
+        factor_rounded : float
+            The normalization factor, rounded to the number of significant digits based on the standard deviation of
+            the integration estimate.
+
+        """
```
ColmTalbot, line 521: Thanks for adding this docstring, the argument description is nice, since we're adding can I suggest also adding a notes section with a [seealso](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-seealso) directive to point to `Halton` and a [versionchanged](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionchanged) directive set to 2.4.0.

ColmTalbot (commented):
This is in great shape, just two last documentation based comments.

bilby/core/prior/dict.py
JasperMartins: I hope I did this in the right way :)

ColmTalbot: I think you could do, e.g., `:func:scipy.integrate.qmc_quad`, but I'm not sure how to link to the external package. I would probably go for just :code:`scipy.integrate.qmc_quad`. The "Scipy's `scipy....`" is kind of redundant and can just be ":code:`scipy....`"

This is how it currently renders (you can download the produced as a CI artefact if you can't build the docs locally.)

image

JasperMartins: Linking to external packages is offered with the sphinx extension `sphinx.ext.intersphinx`. With this extension, one can define where the documentation for external libraries can be found, eg.
```python
intersphinx_mapping = {
    'scipy': ('https://docs.scipy.org/doc/scipy/', None),
}
```
somewhere in the `conf.py`.
Then, using 
```
:py:class:`scipy.stats.qmc.Halton` 
```
will find the right doc page and provide a weblink. Alternatively, I could also just provide a hyperlink inline, but that might break more easily with changes upstream...

ColmTalbot: The intersphinx solution sounds great! To avoid having this PR drag out longer. I would suggest holding that back for another PR. In the meantime, I think update the docstring so that it uses, e.g., :py:class:`scipy.stats.qmc.Halton`. It sounds like when we then add `intersphinx` it will magically update to point to the correct place, in the meantime, I think most people will know what it means.

JasperMartins: I did a test, and it indeed magically appeared when I included the extension

ColmTalbot (approved):
Thanks for pushing this through!

JasperMartins:
Sorry, while taking a last look, I actually caught a bug. The logic enabling marginalization over joint prior keys was not working quite right. I switched to draft so that this is not merged yet. I'll do one last push shortly with a new test case that checks for JointPriors.

JasperMartins:
> I'll do one last push shortly with a new test case that checks for JointPriors.

This turned out to be quite the deep dive. I have implemented two test cases with multivariate Gaussian priors, both with  covariance matrices equal to identity. One case has three dimensions and one case two dimensions, and both cases are constrained to the unit circle in the two dimensions shared- which means they should integrate to the same normalization constant (`1/scipy.stats.chi^2(df=2).cdf(1)`, in fact).

However, using Halton sequences, the integration of the two-dimensional case consistently failed, while the three-dimensional case worked fine. I have also tested the scipy native `qmc_quad` routine, yielding the same, wrong result.
I have played around quite a bit, tried different sizes of the sampling chunk (sizes with $2^m + 1$ for some integer m yielded the right result, actually), tried Sobol sequences (which also yielded wrong results), checked if the problem is the seeding...and tried just standard MC sampling on the unit-cube, which worked consistently. 

While trying to figure out what was going wrong, I also noticed this paragraph in the [scipy documentation of the qmc module](https://docs.scipy.org/doc/scipy/reference/stats.qmc.html):
> Also, to see an improvement over IID MC, (R)QMC requires a bit of smoothness of the integrand, roughly the mixed first order derivative in each direction,[...], must be integral. For instance, a function that is 1 inside the hypersphere and 0 outside of it has infinite variation in the sense of Hardy and Krause for any dimension d > 2.

The nature of the integration we want to perform is just that: we always integrate a characteristic function that is 0 somewhere and 1 elsewhere. 

For these reasons, using QMC appears to be a bad idea after all, although I find it concerning that a standard scipy integration scheme yields (quite) wrong results for such a simple case. I will open an Issue upstream when I find time.

Anyway, the changes to the termination of the integration still seem worthwhile to me, and the added test cases are certainly nice to have. I also think it is still worthwhile to have two integration methods, one "from_samples" and one "from_rescale"), which can be used to check the consistency of the methods. The new commits adopt these changes.

I'm sorry this was only unearthed so late in this pull request. I wonder if this should be moved to an entirely new PR, as this one is related so much to QMC integration.

JasperMartins:
Another update from me: It turns out that the reason for the issue I have described above arises because the rescaling step of the multivariate Gaussian prior reveals the correlations between points produced by the Halton random number generator! Consider the following example:
```python
import bilby
import scipy
import numpy as np

prior_dist = bilby.prior.MultivariateGaussianDist(
    names=["x", "y"],
    mus=[0, 0],
    sigmas=[1, 1],
)


joint_prior_x = bilby.prior.JointPrior(dist=prior_dist, name="x")
joint_prior_y = bilby.prior.JointPrior(dist=prior_dist, name="y")
prior_dict = bilby.prior.PriorDict(
    {
        "x": joint_prior_x_2,
        "y": joint_prior_y_2,
    },
)
```
First, let's produce samples using normal monte carlo:
```
keys = ("x", "y")
sampling_chunk = 2**14
samples = np.random.random(size=(len(keys), sampling_chunk))
samples = np.array(prior_dict.rescale(keys=keys, theta=theta)).reshape((len(keys), sampling_chunk))
samples = {key: np.array(samps) for key, samps in zip(keys, samples)}
```
If we look at a corner plot, everything looks good:
image

Now, lets use Halton:
```python
keys = ("x", "y")
sampling_chunk = 2**14
qrng = scipy.stats.qmc.Halton(d=len(keys), scramble=True)
theta = qrng.random(sampling_chunk).T
samples = np.array(prior_dict.rescale(keys=keys, theta=theta)).reshape((len(keys), sampling_chunk))
samples = {key: np.array(samps) for key, samps in zip(keys, samples)}
```
And the result is clearly broken:
image


What led to considerable confusion on my part during debugging is that `sampling_chunk = 2**14 + 1`, as well as other, seemingly arbitrary changes, works just fine:
image

Also, one can observe how much smoother, i.e. more evenly sampled, the *working* Halton case is compared to normal monte carlo random numbers - indicating advantages for integration that could be exploited.
However, In the end, I think it is more important for the normalization factor integration to be robust to various ways to define priors, rather than squeezing out every bit of efficiency, so my above conclusion still stands.

JasperMartins:
I hope to have a final update on this bug: I have found the bug that led to the correlations in the Quasi-Monte Carlo rescaling step. 

```python
    def _integrate_normalization_factor_from_qmc_quad(self, keys, sampling_chunk):
        qrng = Halton(len(keys), seed=random.rng)
        theta = qrng.random(sampling_chunk).reshape((len(keys), -1)) <-----1.  
        samples = self.rescale(keys=keys, theta=theta)
        samples = np.array({key: samps for key, samps in zip(keys, samples)}).reshape((len(keys, -1)) <----- 2.  
        factor = np.mean(self.evaluate_constraints(samples))
        return factor
```

1. This line must have .T, not .reshape((len(keys,-1)) as the latter leads to a wrong order of the elements. For true Monte Carlo, the difference does not matter, because all entries of theta are independent random variables. For Quasi Monte Carlo, however, the order matters.
2. This line was troubled by issues ultimately solved by #863, hence, this PR now also depends on that PR.

> Also, one can observe how much smoother, i.e. more evenly sampled, the working Halton case is compared to normal monte carlo random numbers - indicating advantages for integration that could be exploited.
However, In the end, I think it is more important for the normalization factor integration to be robust to various ways to define priors, rather than squeezing out every bit of efficiency, so my above conclusion still stands.

I would leave it up to you to decide whether to use Quasi Monte Carlo or "normal" Monte Carlo. In regard to the failing test, conversion_function and so on now require casts to numpy arrays to deal with lists of samples. See https://github.com/bilby-dev/bilby/pull/863 where I have started a discussion on the matter.

JasperMartins:
@ColmTalbot Would there still be interest in (parts) of this pull request? I'd resubmit a reduced version that only implements a relative-error-based termination criterion for the normalization factor estimate without the rescale changes that were necessary for the QMC-based integration.

PR #842, opened 2024-10-31
ENH: implement time marginalization into multiband likelihood

SMorisaki:
This MR implements time marginalization into multiband likelihood. Time marginalization for multiband may be useful since the slow down factor of time marginalization is not as large as that for ROQ or relative binning. The plots below show the likelihood evaluation time of multiband/standard likelihood and with/without time marginalization for 16s XPHM, 64s XPHM, and 128s Pv2_NRtidalv2. The slow down is less than 2 for the first 2 cases.

16s_xphm

64s_xphm_v2

128s_pv2nrtidalv2

### DIFF
--- a/test/gw/likelihood/marginalization_test.py
+++ b/test/gw/likelihood/marginalization_test.py
@@ -165,7 +165,7 @@ class TestMarginalizations(unittest.TestCase):
     The `time_jitter` parameter makes this a weaker dependence during sampling.
     """
     _parameters = product(
-        ["regular", "roq", "relbin"],
+        ["regular", "roq", "relbin", "multiband"],
         ["luminosity_distance", "geocent_time", "phase"],
         [True, False],
         [True, False],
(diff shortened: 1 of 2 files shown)


test/gw/likelihood/marginalization_test.py
```diff
@@ -268,6 +268,18 @@ def setUp(self):
...
         )
 
+        self.multiband_waveform_generator = bilby.gw.WaveformGenerator(
+            duration=self.duration,
+            sampling_frequency=self.sampling_frequency,
+            frequency_domain_source_model=bilby.gw.source.binary_black_hole_frequency_sequence,
+            start_time=1126259640,
+            waveform_arguments=dict(
+                reference_frequency=20.0,
+                minimum_frequency=20.0,
```
ColmTalbot, line 278: I think this line needs to be removed as the `binary_black_hole_frequency_sequence` doesn't use that kwarg.

```suggestion
```

ColmTalbot (approved):

mj-will (approved):

Issue #843, opened 2024-10-31
Add devcontainer file


ColmTalbot:
Now we're on github it would be nice to have at least one predefined devcontainer, possibly one with and without GW requirements.

There is generic documentation on devcontainers [here](https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/adding-a-dev-container-configuration/introduction-to-dev-containers).

Looking for volunteers!

PR #845, opened 2024-10-31
#843 - Added devcontainers

pvcasillasg:
Hello!

I have created the devcontainers using the libraries from conda-forge, as indicated in your installation manual. If you want me to modify it to install in a different way, please let me know.

It’s also possible that some Python libraries may be missing.

I ran a couple of tests on the GW devcontainer, and the example supernova_example.py ran without any issues.

### DIFF
--- a/.devcontainer/gw-container/Dockerfile
+++ b/.devcontainer/gw-container/Dockerfile
@@ -1,0 +1,8 @@
+FROM continuumio/anaconda3
+
+RUN conda install -c conda-forge bilby gwpy python-lalsimulation pymultinest
+
+
+WORKDIR /workspace
+
+COPY ../../. /workspace
--- a/.devcontainer/gw-container/devcontainer.json
+++ b/.devcontainer/gw-container/devcontainer.json
@@ -1,0 +1,13 @@
+{
+  "name": "Bilby GW Development",
+  "build": {
+    "dockerfile": "Dockerfile"
+  },
+  "settings": {
+    "python.pythonPath": "/opt/conda/bin/python",
+    "python.linting.enabled": true,
+    "python.formatting.provider": "black"
+  },
+  "extensions": ["ms-python.python", "ms-toolsai.jupyter"],
+  "postCreateCommand": "/bin/bash"
+}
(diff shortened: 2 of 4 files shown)


pvcasillasg:
Oh, btw, since im using the continuumio/anaconda3 container image as base, it's a quite big image, it will be a great idea to create a registry for this, and just lay the build process on a pipeline, and modify the devcontainer.json

mj-will:
Hi @pvcasillasg, thanks for making a start on this!

I've added some initial comments but @ColmTalbot may have some more comments since he's managed the containers we use in the CI.

.devcontainer/gw-container/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM continuumio/anaconda3
+
+RUN conda install -c conda-forge bilby gwpy python-lalsimulation pymultinest
```
mj-will, line 3: We may want to use a .yml file to configure the environments, what do you think?

Also, I think the [`gw_requirements.txt`](https://github.com/bilby-dev/bilby/blob/main/gw_requirements.txt) file includes the minimum dependencies we need for the GW dockerfile, though we may wish to add more, e.g. `pytest` and the other dependencies we use for testing.

mj-will:
> Oh, btw, since im using the continuumio/anaconda3 container image as base, it's a quite big image, it will be a great idea to create a registry for this, and just lay the build process on a pipeline, and modify the devcontainer.json

We have something along those lines implemented for the CI containers, see https://github.com/bilby-dev/bilby/blob/main/.github/workflows/build-containers.yml

.devcontainer/gw-container/Dockerfile
pvcasillasg: Seems a good solution, i will have a look over this today at afternoon.

Thanks about the context with gw_requirements, i will take that on mind, i will try to provide you a easy solution for add and modify dependencies in case it's necessary.

pvcasillasg:
> > Oh, btw, since im using the continuumio/anaconda3 container image as base, it's a quite big image, it will be a great idea to create a registry for this, and just lay the build process on a pipeline, and modify the devcontainer.json
> 
> We have something along those lines implemented for the CI containers, see https://github.com/bilby-dev/bilby/blob/main/.github/workflows/build-containers.yml

Oh, that's nice, i will try to develop a solution for upload the docker image on the registry.

PR #847, opened 2024-11-01
REFACTOR: make multiband likelihood call Interferometer.get_detector_response

ColmTalbot:
See https://git.ligo.org/lscsoft/bilby/-/merge_requests/1379

I think this has a conflict with https://github.com/bilby-dev/bilby/pull/842.

Based on ongoing discussion about implementing different detector responses, it would be nice for the likelihoods to all access the response functions uniformly. The MB likelihood currently has an almost verbatim copy of the Inteferometer.get_detector_response method. This MR provides a minimal change to be able to directly call that method.…response

### DIFF
--- a/bilby/gw/likelihood/multiband.py
+++ b/bilby/gw/likelihood/multiband.py
@@ -715,26 +715,13 @@ class MBGravitationalWaveTransient(GravitationalWaveTransient):
             An object containing the SNR quantities.
 
         """
-        strain = np.zeros(len(self.banded_frequency_points), dtype=complex)
-        for mode in waveform_polarizations:
-            response = interferometer.antenna_response(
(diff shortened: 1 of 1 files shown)


mj-will:
#842 is merged now, so hopefully the conflicts can be resolved.

mj-will:
@ColmTalbot @SMorisaki is this in a state where we can get it into the next release?

mj-will (approved):
This looks good to me. I've checked and AFAICT it's apply the exact same operations.

This has reminded me we don't have great tests for the mutliband likelihood so it would be great if we could add more in future.

Also, my might want to flag this change to reviewers since it will impact a part of the code that is used for some production analyses.

PR #848, opened 2024-11-01
DEV: make sure all priors return float when needed

ColmTalbot:
For a long time, some prior classes have cast inputs to numpy arrays this MR removes that behaviour and adds a test to make sure that sampling and evaluating the CDF do not cast to array.

This caused an error when evaluating the CDF many times and then combining the results.

I also streamlined some other behaviour in applying prior constraints that were originally flagged by @cjhaster.

I noticed that the Fermi-Dirac CDF hadn't been implemented for some reason so I added that and added to the CI test suite.
Finally, there was an issue with sampling from joint distributions where the results could be returned out of order.

Copied from https://git.ligo.org/lscsoft/bilby/-/merge_requests/1285


bilby/core/prior/analytical.py
```diff
@@ -356,16 +358,16 @@ def ln_prob(self, val):
...
-        cdf[np.where(np.logical_and(-self.minimum < val, val < self.minimum))] = 0.5
-        cdf[upper_indices] = 0.5 + norm * np.log(val[upper_indices] / self.minimum)
-        cdf[np.where(self.maximum < val)] = 1
-        return cdf
+        _cdf = (
+            -norm * np.log(abs(val) / self.maximum)
+            * (val <= -self.minimum) * (val >= -self.maximum)
+            + (0.5 + norm * np.log(abs(val) / self.minimum))
+            * (val >= self.minimum) * (val <= self.maximum)
+            + 0.5 * (val >= -self.minimum) * (val <= self.minimum)
```
mj-will, line 367: I think both statements here have changed from `>` to `>=`, is that intentional?

bilby/core/prior/analytical.py
```diff
@@ -1411,19 +1412,34 @@ def ln_prob(self, val):
...
+
+        References
+        ==========
+
+        .. [1] M. Pitkin, M. Isi, J. Veitch & G. Woan, `arXiv:1705.08978v1
+           <https:arxiv.org/abs/1705.08978v1>`_, 2017.
+        """
+        result = (
+            (np.logaddexp(0, -self.r) - np.logaddexp(-val / self.sigma, -self.r))
+            / np.logaddexp(0, self.r)
```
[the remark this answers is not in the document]
ColmTalbot: I don't think there's a specific benefit, I just found it more readable to have all of the terms use the same format.

bilby/core/prior/analytical.py
mj-will: @ColmTalbot just bumping this

ColmTalbot: I'm going to change this back

```suggestion
            + 0.5 * (val > -self.minimum) * (val < self.minimum)
```

mattpitkin:
I've double checked the Fermi-Dirac prior CDF and it gives the correct expected result.

bilby/core/prior/slabspike.py
```diff
@@ -88,11 +88,11 @@ def rescale(self, val):
...
-        intermediate_indices = np.where(np.logical_and(
+        lower_indices = val < self.inverse_cdf_below_spike
+        intermediate_indices = np.logical_and(
             self.inverse_cdf_below_spike <= val,
-            val <= self.inverse_cdf_below_spike + self.spike_height))[0]
-        higher_indices = np.where(val > self.inverse_cdf_below_spike + self.spike_height)[0]
+            val <= (self.inverse_cdf_below_spike + self.spike_height))
+        higher_indices = val > (self.inverse_cdf_below_spike + self.spike_height)
 
         res = np.zeros(len(val))
```
mj-will, line 97: Looking below these changes, there is some slightly janky logic relating to floats, do you think it's worth trying to fix this in this PR or shall we leave that for a future PR?

bilby/core/prior/dict.py
```diff
@@ -844,17 +848,37 @@ def rescale(self, keys, theta):
...
+            if isinstance(self[key], JointPrior) and self[key].dist.distname not in joint:
+                joint[self[key].dist.distname] = [key]
+            elif isinstance(self[key], JointPrior):
+                joint[self[key].dist.distname].append(key)
+        for names in joint.values():
+            values = list()
+            for key in names:
+                values = np.concatenate([values, result[key]])
+            for key, value in zip(names, values):
+                result[key] = value
```
mj-will, line 868: Could you help me understand what this loop is doing? I'm finding the logic of this part a bit hard to follow, it might benefit from a comment.

mj-will (commented):
LGTM, I just have a couple of questions before I approve.

bilby/core/prior/slabspike.py
ColmTalbot: This module was trickier than the rest to remove the special casing, I'd rather draw the line on this PR and come back to this after.

mj-will: Sure thing.

bilby/core/prior/dict.py
asb5468: +1

test/core/prior/prior_test.py
```diff
@@ -369,6 +389,15 @@ def test_cdf_zero_below_domain(self):
                 )
                 self.assertTrue(all(np.nan_to_num(prior.cdf(outside_domain)) == 0))
 
+    def test_cdf_float_with_float_input(self):
+        for prior in self.priors:
+            if (
+                bilby.core.prior.JointPrior in prior.__class__.__mro__
```
asb5468, line 395: Why is this test skipped for JointPriors with no finite upper bound?

asb5468 (changes requested):
Left a few questions in the code. Wondering what github will do when I press "submit review" with "request changes" chosen.

bilby/core/prior/dict.py
mj-will: Okay, I think I follow. Is this tested somewhere?

Either way, I think it would be useful to add a comment explaining this.

PR #849, opened 2024-11-01
ENH: Analytic aligned spin

ColmTalbot:
As pointed out in https://arxiv.org/abs/2305.19907. The numerical method we use for the aligned spin prior breaks down in the extreme high SNR regime. This MR adds an exact version when using the default spin prior.

Copy of https://git.ligo.org/lscsoft/bilby/-/merge_requests/1282

### DIFF
--- a/bilby/gw/prior.py
+++ b/bilby/gw/prior.py
@@ -1,7 +1,9 @@
 import os
 import copy
+from functools import partial
 
 import numpy as np
+from scipy.integrate import quad
 from scipy.interpolate import InterpolatedUnivariateSpline, interp1d
 from scipy.special import hyp2f1
 from scipy.stats import norm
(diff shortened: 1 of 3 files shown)


test/core/prior/prior_test.py
```diff
@@ -558,6 +558,8 @@ def test_normalized(self):
                 domain = np.linspace(-1e2, 1e2, 1000)
             elif isinstance(prior, bilby.core.prior.FermiDirac):
                 domain = np.linspace(0.0, 1e2, 1000)
+            elif isinstance(prior, bilby.gw.prior.AlignedSpin):
+                domain = np.linspace(prior.minimum, prior.maximum, 10000)
```
mj-will, line 562: Could you add a test for the `quad` method?

ColmTalbot: Good idea, I've added a new test instance.

ColmTalbot: Looks like this is failing one of the precision tests, I'll see what I can do about it.

bilby/gw/prior.py
```diff
@@ -481,6 +483,8 @@ class AlignedSpin(Interped):
     This is useful when using aligned-spin only waveform approximants.
 
     This is an extension of e.g., (A7) of https://arxiv.org/abs/1805.10457.
+    For the simple case of uniform in magnitude and orientation priors, we use an
+    analytic form of the PDF.
```
adivijaykumar, line 487: Can we add a reference to the relevant equation in a paper?

test/core/prior/prior_test.py
ColmTalbot: I added a note in the test, I think this is just not something that will get exactly the right answer as the prior has no support at the boundary.

mj-will (approved):

bilby/gw/prior.py
```diff
@@ -502,17 +505,43 @@ def __init__(self, a_prior=Uniform(0, 1), z_prior=Uniform(-1, 1),
...
+                The integrand for the aligned spin (chi) probability density
+                after performing the integral over spin orientation using a
+                delta function identity.
+                """
+                return a_prior.prob(aa) * z_prior.prob(chi / aa) / aa
+
+            xx = np.linspace(chi_min, chi_max, 10000)
+            yy = [
+                quad(integrand, a_prior.minimum, a_prior.maximum, chi)[0]
+                for chi in xx
```
mj-will, line 524: The old version of this had `1e-32` add if the minimum was zero, was there a reason to remove this?

adivijaykumar: I think it is because the integrand has a `1 / spin_mag` term, and then trapz would complain. Since quad is used now, this is not a problem.

adivijaykumar (approved):

PR #851, opened 2024-11-05
REFACTOR: make ROQ likelihood call Interferometer.get_detector_response

ColmTalbot:
Similar to #847, this PR makes the ROQ likelihood call the standard detector response code. This required some refactoring of the roq source model. After this change, it would be possible to entirely remove that source model and pass everything through the `frequency_sequency` functions directly.

This might need some more refactoring to simplify the logic in the source module.

Needs #847


bilby/gw/detector/interferometer.py
```diff
@@ -299,22 +300,34 @@ def get_detector_response(self, waveform_polarizations, parameters, frequencies=
...
+        else:
+            antenna_time = self.reference_time
+
         signal = {}
         for mode in waveform_polarizations.keys():
             det_response = self.antenna_response(
                 parameters['ra'],
                 parameters['dec'],
-                parameters['geocent_time'],
+                antenna_time,
```
SMorisaki, line 330: Similarly, the input time to `self.time_delay_from_geocenter` should be replaced by `antenna_time`.

PR #853, opened 2024-11-07
BUG: issue warning when prior sampling efficiency is below 1e-3.

AlexandreGoettel:
Addresses #817

### DIFF
--- a/bilby/core/prior/dict.py
+++ b/bilby/core/prior/dict.py
@@ -449,7 +449,7 @@ class PriorDict(dict):
                 return
             efficiency = n_valid / float(n_tested)
             if n_tested >= 1e3 and efficiency < 1e-3:
-                logger.warning("Warning: prior sampling efficiency is very low, please verify its validity.")
+                logger.warning("Prior sampling efficiency is very low, please verify its validity.")
                 efficiency_warning_was_issued = True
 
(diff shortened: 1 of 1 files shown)


bilby/core/prior/dict.py
```diff
@@ -441,10 +441,26 @@ def constraint_keys(self):
...
     def sample_subset_constrained(self, keys=iter([]), size=None):
+        efficiency_warning_was_issued = False
+
+        def check_efficiency(n_tested, n_valid):
+            nonlocal efficiency_warning_was_issued
+            if efficiency_warning_was_issued:
+                return
+            efficiency = n_valid / float(n_tested)
+            if n_tested >= 1e3 and efficiency < 1e-3:
+                logger.debug("Warning: prior sampling efficiency is very low, please verify its validity.")
```
mj-will, line 452: I think this should be `logger.warning` rather than `debug`, what do you think?

AlexandreGoettel: Absolutely!

mj-will:
> This overlaps with #839

@AlexandreGoettel @ColmTalbot shall we try to get this in before #839 since that is now marked as a draft?

mj-will (approved):

ColmTalbot (approved):
I'm happy to get this change in.

PR #855, opened 2024-11-08
TST: patch logger in tests

mj-will:
Some of the tests were modifying the logger in place and mocking the `warning` method. This breaks the logger for any tests are run after this test, as seen in https://github.com/bilby-dev/bilby/pull/42. I've changed these test to patch the logger instead.

### DIFF
--- a/test/gw/detector/psd_test.py
+++ b/test/gw/detector/psd_test.py
@@ -1,4 +1,3 @@
-import logging
 import os
 import unittest
 from unittest import mock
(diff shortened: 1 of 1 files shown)


ColmTalbot (approved):
Great catch! I think we should audit our use of mock and see how much of it we can remove.

mj-will:
> Great catch! I think we should audit our use of mock and see how much of it we can remove.

Yeah, I agree. From a quick look, I think some of it definitely can't be removed (and is probably correct usage of mock)

jacobgolomb:
I'll rebase the branch in #42 after this is merged.

adivijaykumar (approved):
Thanks so much for this @mj-will!

PR #861, opened 2024-11-20
Eccentricy and mean anomaly support in bilby

raffienficiaud:
The PR adds support for the parameters `eccentricity` and `mean_per_ano` anomaly. 

The changes have been tested for supporting the model SEOBNRv5EHM which review has recently been completed.
As this is my first contribution, let me know if you need more unit tests (I cannot run them all locally or in this fork).


bilby/gw/conversion.py
```diff
@@ -259,7 +259,7 @@ def convert_to_lal_binary_black_hole_parameters(parameters):
                         )
                         converted_parameters[f"cos_tilt_{idx}"] = 1.0
 
-    for key in ["phi_jl", "phi_12"]:
+    for key in ["phi_jl", "phi_12", "eccentricity", "mean_per_ano"]:
```
adivijaykumar, line 262: Thanks for this PR! My personal preference would be to not have the `eccentricity` and `mean_per_ano` fields in quasicircular PE files, so I would vote for not including these here at all. But I'd also like to hear opinions from others.

adivijaykumar: Naively, I would have thought that eccentric PE with `gwsignal_binary_black_hole` would work even without changes to this line. Am I missing something?

AntoniRamosBuades: I introduced eccentricity and mean anomaly here because with the addition of eccentricity and mean_per_ano in the arguments of the `gwsignal_binary_black_hole` function in source.py, when one tries to run a quasicircular approximant (which does not specify eccentricity and mean_per_ano) Bilby would crash.  Adding them here is a possible fix, but if you have other suggestions we are happy to modify the code

adivijaykumar: Thanks for explaining. I still think there are other, better ways of dealing with this. Here are some in decreasing order of preference:

1. Refactor the current code by adding a` _base_gwsignal_cbc_fd_waveform` function similar to `_base_lal_cbc_fd_waveform`, and make `gwsignal_binary_black_hole` call this function similar to how `lal_binary_black_hole` called `_base_lal_cbc_fd_waveform`. Write another function `gwsignal_eccentric_binary_black_hole` which also calls `_base_gwsignal_cbc_fd_waveform`. This change will also make things easier if e.g. you want to add e.g. tides to one of the waveforms.

2. If for some reason 1 isn't preferred, I think the easiest way to not make bilby crash is to fix `eccentricity` and `mean_per_ano` to zero in your prior.

3. You could also set a default value of `eccentricity` and `mean_per_ano` in the function definition.

I think 1 is by far the most robust option. Tagging @asb5468 @ColmTalbot @mj-will for opinions.

raffienficiaud: Thank you @adivijaykumar for the suggestions: is [this](https://github.com/AntoniRamosBuades/bilby/commit/1b3527cb85892002081194dc7db31fda64d17a81) addressing it? If yes, I'll continue from there and revert the other changes and tests.

Concerning 2/ and 3/, I would like to avoid default parameters as they may lead to silent mistakes.

adivijaykumar: @raffienficiaud Yes I think that addresses it for the most part! You might want to take out the default waveform_kwargs from inside the `_base_*` function and put it into the individual functions.

test/gw/conversion_test.py
```diff
@@ -480,6 +482,24 @@ def test_generate_all_bbh_parameters(self):
             self.expected_bbh_keys,
         )
 
+    def test_generate_all_bbh_default_parameters(self):
```
adivijaykumar, line 485: This should be removed after the functions in source.py are refactored.

bilby/gw/conversion.py
ColmTalbot: Previously (https://github.com/bilby-dev/bilby/issues/793) we had discussed not making further modifications to this function and any new desired functionality should be leveraged by directly using the new waveform generator as a subclass of `bilby.gw.waveform_generator.WaveformGenerator` as it will lead to much less of the spaghetti we currently have to deal with different specific cases.

ColmTalbot (commented):
I'm generally against this change as I think it encourages us to continue kicking the can down the road on using the flexibility intended in gwsignal.

requirements.txt
```diff
@@ -12,3 +12,5 @@ tqdm
 h5py
 attrs
 importlib-metadata>=3.6; python_version < '3.10'
+parameterized
+gwpy
```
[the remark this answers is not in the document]
raffienficiaud: Removed

test/gw/conversion_test.py
raffienficiaud: Removed

requirements.txt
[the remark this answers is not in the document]
raffienficiaud: Reverted

bilby/gw/conversion.py
raffienficiaud: I reverted the changes in this file and the related unit test.

raffienficiaud:
> I'm generally against this change as I think it encourages us to continue kicking the can down the road on using the flexibility intended in gwsignal.

I am not sure to understand this statement @ColmTalbot : are you suggesting that this change is not required as gwsignal is supposed to handle it? There is a [plugin](https://git.ligo.org/waveforms/reviews/lalsuite/-/merge_requests/119) infra coming into `gwsignal` but it will take some time to get it released.

mj-will:
> > I'm generally against this change as I think it encourages us to continue kicking the can down the road on using the flexibility intended in gwsignal.
> 
> I am not sure to understand this statement @ColmTalbot : are you suggesting that this change is not required as gwsignal is supposed to handle it? There is a [plugin](https://git.ligo.org/waveforms/reviews/lalsuite/-/merge_requests/119) infra coming into `gwsignal` but it will take some time to get it released.

There is some more context in this issue: https://github.com/bilby-dev/bilby/issues/793.

FWIW, I agree with Colm. I really think we should avoid adding more waveforms via this approach if possible.

raffienficiaud:
@mj-will @ColmTalbot thank you for your comments. I know the work that has been done on `gwsignal` to make the waveform generator interface more flexible (and I extended it recently). However, this may take some amount of time to get into the main branch (there are discussions to get the review started in the coming weeks).

The same issue as the one raised here affects LALSuite BTW, and we have to make intermediate steps until `gwsignal` branch is properly merged, so that we are on tracks to get the waveform model on prod next year.

Until `gwsignal` changes are merged, what do we do?

mj-will:
@raffienficiaud can this be closed in favour of: https://github.com/bilby-dev/bilby/pull/877?

PR #863, opened 2024-11-22
Conserve rescale shape

JasperMartins:
This PR updates the `rescale` functions of `PriorDict` and `ConditionalPriorDict` to preserve the correct shape of the rescaled samples.

The PR also updates the behavior of `JointPrior.rescale` and `BaseJointPriorDist`. 

`JointPrior.rescale` now returns a *mutable* numpy array. If not all required keys have been requested for rescaling, the arrays is filled with `np.nan`. `BaseJointPriorDist` keeps track of this array. Once all required keys are requested, the rescale operation is performed, and the returned arrays are populated with the rescaled values. 
This change enables out-of-order sets of keys in `rescale` - previously `priordict.rescale(keys=["a", "JointPrior_b", "c", "JointPrior_d"], theta=...)` where "JointPrior_b" and "JointPrior_d" share the same `dist` would have resulted in an order `[*samples_a, *samples_c, *samples_b, *samples_d]`.

Relates to the discussion in https://github.com/bilby-dev/bilby/pull/850

### DIFF
--- a/bilby/core/prior/dict.py
+++ b/bilby/core/prior/dict.py
@@ -600,18 +600,21 @@ class PriorDict(dict):
         ==========
         keys: list
             List of prior keys to be rescaled
-        theta: list
-            List of randomly drawn values on a unit cube associated with the prior keys
+        theta: dict or array-like
+            Randomly drawn values on a unit cube associated with the prior keys
 
         Returns
         =======
-        list: List of floats containing the rescaled sample
+        list:
+            If theta is 1D, returns list of floats containing the rescaled sample.
+            If theta is 2D, returns list of lists containing the rescaled samples.
         """
-        theta = list(theta)
+        theta = [theta[key] for key in keys] if isinstance(theta, dict) else list(theta)
         samples = []
         for key, units in zip(keys, theta):
             samps = self[key].rescale(units)
-            samples += list(np.asarray(samps).flatten())
+            # turns 0d-arrays into scalars
+            samples.append(np.squeeze(samps).tolist())
         return samples
 
     def test_redundancy(self, key, disable_logging=False):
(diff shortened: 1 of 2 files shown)


bilby/core/prior/dict.py
```diff
@@ -600,18 +600,23 @@ def rescale(self, keys, theta):
...
-        theta = list(theta)
+        theta = [theta[key] for key in keys] if isinstance(theta, dict) else list(theta)
         samples = []
         for key, units in zip(keys, theta):
             samps = self[key].rescale(units)
-            samples += list(np.asarray(samps).flatten())
+            samples.append(samps)
+        for i, samps in enumerate(samples):
+            # turns 0d-arrays into scalars
+            samples[i] = np.squeeze(samps).tolist()
```
JasperMartins, line 619: I wonder if it would not make more sense to preserve numpy arrays to make it easier to define conversion_functions and so on without needing to cast to array again

ColmTalbot: I think it is would be worth returning arrays, I think it's worth accepting that this is going to not be fully backward compatible.

bilby/core/prior/dict.py
```diff
@@ -600,18 +600,23 @@ def rescale(self, keys, theta):
...
             List of prior keys to be rescaled
-        theta: list
-            List of randomly drawn values on a unit cube associated with the prior keys
+        theta: dict or array-like
+            Randomly drawn values on a unit cube associated with the prior keys
 
         Returns
         =======
-        list: List of floats containing the rescaled sample
+        list:
```
ColmTalbot, line 608: It might be worth returning the same type as the input, i.e., return a dict if a dict is passed.

Another option would be to add a specific method to handle dicts, I've thought about doing this a few times before, e.g., `PriorDict.rescale_dict` that just wraps `PriorDict.rescale`.

bilby/core/prior/joint.py
```diff
@@ -94,15 +95,24 @@ def filled_rescale(self):
         Check if all the rescaled parameters have been filled.
         """
 
-        return not np.any([val is None for val in self.rescale_parameters.values()])
+        return not np.any([val is None for val in self._rescale_parameters.values()])
+
+    def set_rescale(self, key, values):
+        values = np.array(values)
+        self._rescale_parameters[key] = values
+        self._rescaled_parameters[key] = np.atleast_1d(np.ones_like(values)) * np.nan
```
ColmTalbot, line 103: I can't see where these values are set to a non-trivial value, what is the distinction between these two?

ColmTalbot (commented):
Thanks for opening this, I think this will be a good change, and I'm in favour of also not pushing everything through lists.

I don't really follow the logic in the joint prior, so some additional docstrings/comments would probably help there.

bilby/core/prior/joint.py
JasperMartins: The values are updated in-place in the rescale function:
https://github.com/bilby-dev/bilby/blob/4e1a3d2d50f5abc97a3cb95153b7dac1eefa410b/bilby/core/prior/joint.py#L345-L350

https://github.com/bilby-dev/bilby/blob/4e1a3d2d50f5abc97a3cb95153b7dac1eefa410b/bilby/core/prior/joint.py#L819-L829

So, in `self._rescaled_parameters,` the values of the rescaled parameters are stored once the rescaling operation is performed. Prior to this, the array contains only `np.NaN` to clarify that the rescaling has not yet happened. The per-key numpy arrays stored in `self._rescaled_parameters` are returned as references in `JointPrior.rescale`. Once all keys have been requested, the values in the arrays are updated in place, meaning that they will also be updated in the array returned by `JointPrior.rescale`. 

I agree that this should be documented better, particularly to ensure that `PriorDict.rescale` is not updated in a way that loses track of the reference to the arrays stored in `BaseJointPriorDict._rescaled_parameters` in the future.

bilby/core/prior/dict.py
JasperMartins: This could also apply to the conservation of numpy arrays, ie return a dict, list or array depending on the input.

JasperMartins:
I have updated the changes to joint priors with more comments / docstrings and changed variable names to be more reflective of what happens.

For now, I have left the return-values of `PriorDict.rescale()` and `ConditionalPriorDict.rescale()` as lists. Changing to arrays should, I think, be accompanied by a larger-scale switch to numpy arrays for the return values of methods of `Prior()` that I don't have time for right at the moment.

PR #864, opened 2024-11-22
Draft: Add support for conditional JointPriors

JasperMartins:
This fairly big PR updates ConditionalPrior, ConditionalPriorDict, JointPrior, and MultivariateGaussianDist. Since the changes are highly intertwined, I opted for only one PR. The main goal of this PR is to make it possible to use ConditionalJointPriors. 

The changes are the following:

1.  A bug in `ConditionalPriorDict.rescale` is fixed. Previously, it would not have been possible to rescale a subset of keys. Also, the check if the setting of the reference_params as done for `ConditionalPriorDict.sample` supports array is applied for `rescale` as well. Further, both `ConditonalPriorDict.rescale` and `sample` were updated to no longer need to create a copy of the `ConditionalPriorDict` with the requested subset of keys.
2. `ConditionalPriorDict.check_conditions_resolved` now checks if `JointPrior.dist` dependencies are resolved as well. If a JointPrior is a dependency of another prior, all JointPriors associated with the dist of that JointPrior need to be evaluated before the dependency is evaluated - otherwise, rescaling would always fail.
3. `ConditionalPrior` is updated to support `JointPrior`s. This required accepting the named argument `dist` and a bit more flexibility on which named arguments to pass towards the init-function of the superclass. This also resulted in a change to the method `get_instantiation_dict`, which now only collects properties required to initialize the superclass and not all (named) arguments of `ConditionalPrior`.
4. `JointPrior` is updated to be able to set properties of `BaseJointPriorDist`-subclasses if these have explicit setter methods for these properties. That way, it ensures that no properties are changed in ways that would potentially break `BaseJointPriorDist`-subclasses.
5. `MultivariateGaussianDist` is updated to allow setting the mode to sample/rescale/calculate. This makes it possible to use `ConditionalPriorDict` to have a prior select the current mode. Using multimodal multivariate Gaussians for sampling is more straightforward that way, as it allows modes to "die off" during sampling if they do not contribute to the posterior. The implementation also supports arrays of different modes for different samples. Further, a few changes are made to `_rescale`, `_sample` and `_ln_prob` to be more efficient for large numbers of samples.
6. A bug in `BaseJointPriorDist` is fixed where `bounds` were not applied correctly.
7. Test cases for the introduced functionality are added.


JasperMartins:
Regardless of my above discussion point (which would be a larger overall change), I think the changes are ready for enabling Joint Conditional Priors, so I would like to ask about the prospects of this PR.

PR #865, opened 2024-11-26
CI: specify `needs` for `pypi-publish`

mj-will:
Add `needs` to the `pypi-published` step of the `publish.yml` workflow.

This ensured `pypi-published` runs _after_ `build`.

Close #860

PR #866, opened 2024-11-26
DOC: add instructions for making releases

mj-will:
As discussed on the dev call, I've added a new section to the contributing guidelines describing how to make releases.

This assumes that the CI works, which should be fixed by https://github.com/bilby-dev/bilby/pull/865

I'm happy to move this elsewhere if people think there's a better place for it.

I also fixed https://github.com/bilby-dev/bilby/issues/805 since it was making changes to the file. The issue was the use of relative paths, so I've changed it to just pull the file from the git repo.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -353,6 +353,6 @@ parts of the code. Note that this document is not programmatically generated and
 so may get out of date with time. If you notice something wrong, please open an
 issue.
 
-![bilby overview](docs/images/bilby_layout.png)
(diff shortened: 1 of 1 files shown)


mj-will:
> Is it worth saying something about what to do if the automatic deployment doesn't work? Even if it is just "tag Colm and Michael"?

Good point, I'll add something.

ColmTalbot (approved):

adivijaykumar (approved):

PR #867, opened 2024-11-27
ENH: add cosmology to CBC result

mj-will:
Add `cosmology` to the `CompactBinaryCoalescenceResult`.

If we run analyses with a different cosmology, I think this should be recorded in the result file. My proposal is to add the `cosmology` in a new dictionary within `meta_data` called `global_meta_data`. This is inline with the changes proposed in https://github.com/bilby-dev/bilby/pull/873. I've implemented it following the same pattern used for other properties, such as the waveform arguments.

Whilst making these changes, I realised that for this to work, the encoding/decoding on astropy objects wasn't working correctly for `hdf5` files. So a lot of changes are related to this.

**Changes**

- Add `cosmology` to the meta data dictionary when calling run sampler
- Add `cosmology` to the meta data dictionary when making an instance of `CompactBinaryCoalescenceResult`
- Improve `encode/decode_astropy_cosmology/quantity` to make use of built-in astropy functions and to more closely match the `JsonCustomEncode` implemented in astropy. The `decode` functions should be backwards compatible.
- Add `encode/decode_astropy_unit`
- Add `decode_hdf5_dict` to ensure astropy types are decoded
- Add tests to ensure the result objects with cosmology can be saved to all available formats
- Improved/added doc-strings for various methods

### DIFF
--- a/bilby/core/sampler/__init__.py
+++ b/bilby/core/sampler/__init__.py
@@ -251,6 +251,14 @@ def run_sampler(
     meta_data["likelihood"] = likelihood.meta_data
     meta_data["loaded_modules"] = loaded_modules_dict()
     meta_data["environment_packages"] = env_package_list(as_dataframe=True)
+    # Add the cosmology if available
+    if "cosmology" not in meta_data:
+        try:
+            from bilby.gw.cosmology import get_cosmology
+
+            meta_data["cosmology"] = get_cosmology()
+        except ImportError:
+            logger.info("Could not specify cosmology in meta data")
 
     if command_line_args.bilby_zero_likelihood_mode:
         from bilby.core.likelihood import ZeroLikelihood
--- a/bilby/core/sampler/__init__.py
+++ b/bilby/core/sampler/__init__.py
@@ -251,14 +251,6 @@ def run_sampler(
     meta_data["likelihood"] = likelihood.meta_data
     meta_data["loaded_modules"] = loaded_modules_dict()
     meta_data["environment_packages"] = env_package_list(as_dataframe=True)
-    # Add the cosmology if available
-    if "cosmology" not in meta_data:
-        try:
-            from bilby.gw.cosmology import get_cosmology
-
-            meta_data["cosmology"] = get_cosmology()
-        except ImportError:
-            logger.info("Could not specify cosmology in meta data")
 
     if command_line_args.bilby_zero_likelihood_mode:
         from bilby.core.likelihood import ZeroLikelihood
--- a/bilby/core/utils/io.py
+++ b/bilby/core/utils/io.py
@@ -11,7 +11,6 @@ import numpy as np
 import pandas as pd
 
 from .log import logger
-from .introspection import infer_args_from_method
 
 
 def check_directory_exists_and_if_not_mkdir(directory):
@@ -87,11 +86,7 @@ class BilbyJsonEncoder(json.JSONEncoder):
 
 
 def encode_astropy_cosmology(obj):
-    cls_name = obj.__class__.__name__
-    dct = {key: getattr(obj, key) for key in infer_args_from_method(obj.__init__)}
-    dct["__cosmology__"] = True
-    dct["__name__"] = cls_name
-    return dct
+    return {"__cosmology__": True, **obj.to_format("mapping")}
 
 
 def encode_astropy_quantity(dct):
@@ -103,11 +98,10 @@ def encode_astropy_quantity(dct):
 
 def decode_astropy_cosmology(dct):
     try:
-        from astropy import cosmology as cosmo
+        from astropy.cosmology import Cosmology
 
-        cosmo_cls = getattr(cosmo, dct["__name__"])
-        del dct["__cosmology__"], dct["__name__"]
-        return cosmo_cls(**dct)
+        del dct["__cosmology__"]
+        return Cosmology.from_format(dct, format="mapping")
     except ImportError:
         logger.debug(
             "Cannot import astropy, cosmological priors may not be " "properly loaded."
@@ -241,6 +235,8 @@ def decode_from_hdf5(item):
             output = item
     elif isinstance(item, np.bool_):
         output = bool(item)
+    elif isinstance(item, dict) and "__cosmology__" in item:
+        output = decode_astropy_cosmology(item)
     else:
         output = item
     return output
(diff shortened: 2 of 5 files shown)


mj-will:
I've updated the PR to added support for saving astropy cosmologies to our `HDF5` files.

Whilst doing this, also updated the code to use `to_format`/`from_format` for encoding and encoding the cosmologies sine it's the recommended way https://docs.astropy.org/en/stable/cosmology/io/builtin.html#cosmology-io-builtin

bilby/core/sampler/__init__.py
```diff
@@ -251,6 +251,14 @@ def run_sampler(
     meta_data["likelihood"] = likelihood.meta_data
     meta_data["loaded_modules"] = loaded_modules_dict()
     meta_data["environment_packages"] = env_package_list(as_dataframe=True)
+    # Add the cosmology if available
```
ColmTalbot, line 254: I'd rather not add this here, I think it sets a bad precedent of adding application-specific logic into core. Is there some reason adding this to the `CBCResult` is not sufficient? I think "users didn't specify the result type" is not a sufficient reason here.

mj-will: I agree that 'users didn't specify the result type' shouldn't be reason enough in principle, but I'm slightly concerned the documentation is sufficiently clear on this (AFAICT, only two of all our GW examples specify it). Therefore, anyone that's used one of the examples that doesn't have it as a starting point may not have specified the class. It's also different to e.g. the likelihood metadata which is always saved irrespective of the results class, meaning it can be recovered even if the wrong result class is used.

I do see your point, as this is clearly GW code. This may be out-of-scope for this MR, but I could imagine having an object that keeps track of some of the global variables, e.g. `rng`? This could then be saved in the metadata. If we had something like this, calling `set_cosmology` could just add the cosmology to that object?

Either way, I think we should update the GW examples to include the correct result class.

ColmTalbot: I agree with all of this, especially updating examples (and docs). I think some kind of generic metadata object in the background is a good idea, adding to this in `set_cosmology` also sounds great. Another thing we can think about is setting a default cosmology via environment variable, I think astropy does similar, we could even piggyback off theirs.

mj-will: So I've removed these lines and had a go at implementing a global meta data object in: https://github.com/bilby-dev/bilby/pull/873. I've also opened an issue to track updates to the examples: https://github.com/bilby-dev/bilby/pull/873

mj-will: I've also moved `cosmology` to the `global_meta_data` so it is consistent with #873

bilby/core/utils/io.py
```diff
@@ -264,6 +352,13 @@ def encode_for_hdf5(key, item):
     """
     from ..prior.dict import PriorDict
 
+    try:
+        from astropy import cosmology as cosmo, units
```
ColmTalbot, line 356: I'm not wild about having this import here, I think it could be avoided, but since it is only a debug message I'm happy to let it slide for a while.

_unrelated_: We could consider changing this function to use multiple dispatch at some point down the road, that would avoid ugly import tests and also allow downstream users to add their own encoding/decoding.

ColmTalbot (approved):

bilby/core/utils/io.py
mj-will: Yeah, I agree. At the time, I couldn't come up with a clean alternative.

On multiple dispatch: that's an interesting option, might be something to bring up on a dev call.

mj-will:
Following from the discussion on the last dev call, I checked some existing files.

For GWTC-3 and GWTC-4 files it seems to work:

```python
>>> import bilby
>>> bilby.gw.result.CBCResult.from_json("/home/pe.o3/o3a_final/run_directories/S190706a/ProdF9/result/ProdF9_data0_1246487219-346191_analysis_H1L1V1_dynesty_merge_result.json")
<bilby.gw.result.CompactBinaryCoalescenceResult object at 0x7fb556a93ed0>
>>> bilby.gw.result.CBCResult.from_hdf5("/home/pe.o4/GWTC4/working/S230608as/bilby-IMRPhenomXPHM-SpinTaylor/final_result/bilby-IMRPhenomXPHM-SpinTaylor_data0_1370292665-161092_analysis_H1L1_merge_result.hdf5")
<bilby.gw.result.CompactBinaryCoalescenceResult object at 0x7fb5324a7f10>
```
However, for a bilby GWTC-1 file, it's current breaking but works with 2.4.0. I'll see if I can figure out why and fix it.

mj-will:
> Following from the discussion on the last dev call, I checked some existing files.
> 
> For GWTC-3 and GWTC-4 files it seems to work:
> 
> ```python
> >>> import bilby
> >>> bilby.gw.result.CBCResult.from_json("/home/pe.o3/o3a_final/run_directories/S190706a/ProdF9/result/ProdF9_data0_1246487219-346191_analysis_H1L1V1_dynesty_merge_result.json")
> <bilby.gw.result.CompactBinaryCoalescenceResult object at 0x7fb556a93ed0>
> >>> bilby.gw.result.CBCResult.from_hdf5("/home/pe.o4/GWTC4/working/S230608as/bilby-IMRPhenomXPHM-SpinTaylor/final_result/bilby-IMRPhenomXPHM-SpinTaylor_data0_1370292665-161092_analysis_H1L1_merge_result.hdf5")
> <bilby.gw.result.CompactBinaryCoalescenceResult object at 0x7fb5324a7f10>
> ```
> 
> However, for a bilby GWTC-1 file, it's current breaking but works with 2.4.0. I'll see if I can figure out why and fix it.

I've fixed this is my last commit. The issue was that the `__cosmology__` entry had already been deleted before falling back to the legacy decoding causing the `del` there to fail.

GregoryAshton:
@mj-will overall this looks like a good. But, it does add a lot of overhead and tying into astropy. Is there any risk that we will end up unable to read/write results because of an astropy version change? E.g., if an astropy cosmology gets added/changed is this robust?

mj-will:
> @mj-will overall this looks like a good. But, it does add a lot of overhead and tying into astropy. Is there any risk that we will end up unable to read/write results because of an astropy version change? E.g., if an astropy cosmology gets added/changed is this robust?

@GregoryAshton I agree this does increase our dependence on astropy but, to me, this makes us more robust to astropy changes since we're using their own API to read/write objects rather than relying on introspection. The `to_format`/`from_format` methods have been around since v5 so I don't expect they will have large changes without a fair bit of warning. We could however implement a fallback for if the encoding fails, though I'm not quite sure what the best option would be (I'm not sure what exception one could catch). Do you have any thoughts?

As for overhead, other than the initial import, I'd hope most of the operations are pretty lightweight. Is there a particular call/operation you know to be slow?

mj-will:
Thanks, @GregoryAshton, given that are you happy to approve this?

ColmTalbot (approved):

GregoryAshton (approved):

PR #868, opened 2024-11-27
ENH: add cosmology to `CBCPriorDict`

mj-will:
Add attributes and a method that allow the user to check if a `CBCPriorDict` contains any cosmological priors and check they use the same cosmology.

This should simplify adding a cosmology to `bilby_pipe` as it will allows to easily check the priors are valid and then pull the cosmology.

### DIFF
--- a/test/gw/prior_test.py
+++ b/test/gw/prior_test.py
@@ -174,11 +174,11 @@ class TestBBHPriorDict(unittest.TestCase):
         self.bbh_prior_dict["luminosity_distance"] = bilby.gw.prior.UniformComovingVolume(
             minimum=10, maximum=10000, name="luminosity_distance"
         )
-        self.assertTrue(self.bbh_prior_dict.is_cosmological())
+        self.assertTrue(self.bbh_prior_dict.is_cosmological)
 
     def test_is_cosmological_false(self):
         del self.bbh_prior_dict["luminosity_distance"]
-        self.assertFalse(self.bbh_prior_dict.is_cosmological())
+        self.assertFalse(self.bbh_prior_dict.is_cosmological)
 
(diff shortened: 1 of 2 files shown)


ColmTalbot (approved):

test/gw/prior_test.py
```diff
@@ -170,6 +170,44 @@ def test_add_constraint_prior_not_redundant(self):
...
+            ["luminosity_distance", "redshift"],
+        )
+        self.assertRaises(ValueError, self.bbh_prior_dict.check_valid_cosmology)
+
+    def test_cosmology(self):
+        self.bbh_prior_dict["luminosity_distance"] = bilby.gw.prior.UniformComovingVolume(
+            minimum=10, maximum=10000, name="luminosity_distance"
+        )
+        self.assertEqual(
+            self.bbh_prior_dict.cosmology,
```
adivijaykumar, line 207: Is there a good reason not to use `cosmology_equal` from `astropy` here?

mj-will: Good point, I don't think there is. I'll update it.

mj-will: Actually, on second thought, using `assertEqual` gives a more useful error message. `assertTrue` just gives `AssertionError: False is not true`

adivijaykumar: Actually, seems like you can pass a custom message to `assertTrue`, for instance: `self.assertTrue(cosmology_equal(cosmo_1, cosmo_2), "The two cosmologies are not equivalent")`. Your call if you want to put this in though :)

mj-will: I think I'll leave it as is, the current version prints the cosmologies, meaning one can compare the values, which I think is useful.

PR #870, opened 2024-12-03
[BUG]: fixing dtypes being changed in plot_interferometer_waveform_posterior

foocheng:
Fixing a small bug in CBCResult. plot_interferometer_waveform_posterior, where the pandas.posterior.iterrows() function sometimes changes dtype of input (e.g. from float to complex), leading to some errors later down the line. Replaced with a very similar method itertuples which preserves dtypes of inputs.

### DIFF
--- a/bilby/gw/result.py
+++ b/bilby/gw/result.py
@@ -496,8 +496,8 @@ class CompactBinaryCoalescenceResult(CoreResult):
 
         fd_waveforms = list()
         td_waveforms = list()
-        for _, params in samples.iterrows():
-            params = dict(params)
+        for params in samples.itertuples(index=False):
+            params = dict(params._asdict())
             wf_pols = waveform_generator.frequency_domain_strain(params)
             fd_waveform = interferometer.get_detector_response(wf_pols, params)
             fd_waveforms.append(fd_waveform[frequency_idxs])
--- a/bilby/gw/result.py
+++ b/bilby/gw/result.py
@@ -496,8 +496,7 @@ class CompactBinaryCoalescenceResult(CoreResult):
 
         fd_waveforms = list()
         td_waveforms = list()
-        for params in samples.itertuples(index=False):
-            params = dict(params._asdict())
+        for params in samples.to_dict(orient="records"):
             wf_pols = waveform_generator.frequency_domain_strain(params)
             fd_waveform = interferometer.get_detector_response(wf_pols, params)
             fd_waveforms.append(fd_waveform[frequency_idxs])
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -16,6 +16,7 @@ Bruce Edelman
 Carl-Johan Haster
 Cecilio Garcia-Quiros
 Charlie Hoy
+Cheng Foo
 Chentao Yang
 Christopher Philip Luke Berry
 Christos Karathanasis


adivijaykumar:
@foocheng thanks for this PR! Would it be possible for you to illustrate one case in which the current method fails? I am having trouble reproducing this behaviour.

foocheng:
Hi @adivijaykumar, for some reason I'm only seeing this behaviour with result files from parallel_bilby runs rather than bilby_pipe runs, though in principle the resulting parameters look similar. I'm not quite sure why there should be differences in processing the parallel_bilby results here vs. the bilby_pipe ones, perhaps it's somehow related to [this issue](https://github.com/bilby-dev/bilby/issues/871#issuecomment-2536455393)) I've raised? But regardless, the iterrows function doesn't preserve dtypes.

Here is a minimal working example (similar to the one I've just put on [this issue](https://github.com/bilby-dev/bilby/issues/871#issuecomment-2536455393)):

### Minimal Working Example: 
**Creating a parallel_bilby generated result file** (similar to [this minimal example](https://git.ligo.org/lscsoft/bilby_pipe/-/tree/master/examples/injection?ref_type=heads ) for bilby_pipe, but for parallel_bilby). 
Rename this to`bbh_injection_parallel.ini`
[bbh_injection_parallel.txt](https://github.com/user-attachments/files/18098934/bbh_injection_parallel.txt)
Rename this to `bbh_injection_prior.prior`
[bbh_injection_prior.txt](https://github.com/user-attachments/files/18098850/bbh_injection_prior.txt)

Run the following command to create injection data 
`bilby_pipe_create_injection_file bbh_injection_prior.prior --n-injection 10 --generation-seed 1234 -f injections.json`

Create result file with parallel_bilby:
`parallel_bilby_generation bbh_injection_parallel.ini --submit`

Then, when trying to call the results, adding print statements before and after line 479 here: 
https://github.com/bilby-dev/bilby/blob/cccdf891e82d46319e69dbfdf48c4970b4e9a727/bilby/gw/result.py#L479
we see that float dtypes are converted to complex dtypes after the iterrows operation. However, with the merge request change here, this dtypes should be conserved. 

So 
`from bilby.gw.result import CBCResult`
`test_hdf5 = CBCResult.from_hdf5("outdir_bbh_injection/result/bbh_injection_0_result.hdf5")`
And then 
```
ifos = bilby.gw.detector.InterferometerList(["H1", "L1"])
ifos.set_strain_data_from_zero_noise(
    duration=duration,
    sampling_frequency=4096,
    start_time=injection_parameters["geocent_time"] - 2,
)

for ifo in ifos:
    test.plot_interferometer_waveform_posterior(
            interferometer=ifo, n_samples=500, save=False
        )
    plt.show()
    plt.close()
```

And the print statements should show. Note that unlike [this issue](https://github.com/bilby-dev/bilby/issues/871#issuecomment-2536455393)), this happens regardless if you load the result in from hdf5 or from json, so it shouldn't be because of the Result class type. 

This won't affect most use cases, since then casting Complex values to real will just discard the complex part (which is just 0.0j). However, this makes it incompatible with some functions such as `pyseobnr.generate_waveform.GenerateWaveform, which have checks in place to make sure certain parameters such as distance are real numbers.

bilby/gw/result.py
```diff
@@ -476,8 +476,8 @@ def plot_interferometer_waveform_posterior(
 
         fd_waveforms = list()
         td_waveforms = list()
-        for _, params in samples.iterrows():
-            params = dict(params)
+        for params in samples.itertuples(index=False):
+            params = dict(params._asdict())
```
mj-will, line 480: I think `dict` is redundant as of version Python 3.8 here, see [the docs here](https://docs.python.org/3/library/collections.html#collections.somenamedtuple._asdict).

adivijaykumar (approved):
Looks good to me, save for the change that Michael suggested.

bilby/gw/result.py
ColmTalbot: I think we can instead do the following and avoid the explicit cast to dict.

```python
        for params in samples.to_dict(orient="records"):
```

mj-will: Ah, that's much neater.

mj-will: @foocheng could you update the PR to use Colm's suggestion?

adivijaykumar (approved):

bilby/gw/result.py
mj-will: Great, thanks!

mj-will:
Rebasing so we can merge this.

PR #872, opened 2024-12-05
MAINT: dynesty - reduce number of calls to add_live_points

ColmTalbot:
A bug in dynesty (https://github.com/joshspeagle/dynesty/issues/490) means that the chain can be corrupted if we write a checkpoint after catching an interrupt.

Currently, we call the unsafe method (add_live_points) more frequently than we need to. This PR makes it so that we only call this when making trace plots and after writing our main checkpoint file. This should reduce the number of calls drastically and practically eliminate our exposure to the issue.

Longer term, we may want to think about moving away from writing a new checkpoint file when exiting and rely on previously written checkpoints as pointed out in the linked issue, writing a checkpoint on exit will always be fundamentally unsafe, even if it can be made practically safe.

### DIFF
--- a/bilby/core/sampler/dynesty.py
+++ b/bilby/core/sampler/dynesty.py
@@ -667,7 +667,7 @@ class Dynesty(NestedSampler):
 
     def finalize_sampler_kwargs(self, sampler_kwargs):
         sampler_kwargs["maxcall"] = self.n_check_point
-        sampler_kwargs["add_live"] = True
+        sampler_kwargs["add_live"] = False
 
     def _run_external_sampler_with_checkpointing(self):
         """
@@ -675,8 +675,11 @@ class Dynesty(NestedSampler):
         periods of time (less than the checkpoint time) and if sufficient
         time has passed, write a checkpoint before continuing. To get the most
         informative checkpoint plots, the current live points are added to the
-        chain of nested samples within dynesty and have to be removed before
-        restarting the sampler.
+        chain of nested samples before making the plots and have to be removed
+        before restarting the sampler. We previously used the dynesty internal
+        version of this, but this is unsafe as dynesty is not capable of
+        determining if adding the live points was interrupted and so we want to
+        minimize the number of times this is done.
         """
 
         logger.debug("Running sampler with checkpointing")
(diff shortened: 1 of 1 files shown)


GregoryAshton:
@ColmTalbot it looks like the CI is failing because dynamic doesn't have the `add_live` method - should this only apply to standard dynesty?

bilby/core/sampler/dynesty.py
```diff
@@ -706,15 +708,27 @@ def _run_external_sampler_with_checkpointing(self):
...
                 self.write_current_state()
+                self._add_live()
                 self.plot_current_state()
-        if getattr(self.sampler, "added_live", False):
-            self.sampler._remove_live_points()
+                self._remove_live()
 
+        self._remove_live()
+        if "add_live" in sampler_kwargs:
+            sampler_kwargs["add_live"] = True
```
mj-will, line 717: Could you help me understand why we need these lines?

Following the call to `finalize_sampler_kwargs`, `add_live` should always be false. Is this reverting that? If so, what if `add_live` was `False` to begin with?

ColmTalbot: We run dynesty many times in short bursts to enable our checkpointing. If we specify `add_live=True` the live points are added every time, which can lead to bad behaviour. But we need to add the live points in the final pass through run_nested so that they are added for the final posterior.

ColmTalbot: Oh, you mean if the user specified `add_live=False`. This would break that behaviour, I guess I should take this from the original kwargs.

mj-will: Yes, that's what I meant. I think it from the original kwargs makes sense.

mj-will (commented):
Following the change to `added_live`, the changes look good to me. 

@ColmTalbot have you had a chance to test this via `bilby_pipe`? I appreciate the tests should cover this but I think it would be good to be sure. What do you think @GregoryAshton ?

GregoryAshton:
All looks good to me too.

GregoryAshton (approved):

mj-will (approved):

PR #873, opened 2024-12-06
ENH: add global meta data

mj-will:
Based on the discussion in https://github.com/bilby-dev/bilby/pull/867, I've made a start on a possible implementation of a global meta data object.

For now, it's just a singleton dictionary that, by default, tracks the `rng`. Things like the `cosmology` can then be added when they're set.

**Motivation**

This change makes it easier to keep track of any global variables and allows us to ensure they're saved in the result file after running a sampler.

**Changes** (updated 27/01/25)

- Add a new `meta_data` submodule and `GlobalMetaData` class
- Add functions for encoding/decoding numpy random generators
- Add test for saving generators
- Extend `seed`, `set_cosmology` and `_set_default_cosmology` to update the global meta data when called.
- Add functions for encoding `UserDict` and `UserList` subclasses. These rely on the fact both classes have a `data` attribute, see [the Python docs](https://docs.python.org/3/library/collections.html#userdict-objects)

**Testing**

These changes are mostly covered by the existing tests save result files, but I added a test for the encoding/decoding change and for the global meta data class.

### DIFF
--- a/bilby/core/utils/random.py
+++ b/bilby/core/utils/random.py
@@ -14,8 +14,8 @@ def seed(seed):
     from .meta_data import global_meta_data
 
     Generator.rng = default_rng(seed)
-    global_meta_data.add_to_meta_data("rng", Generator.rng)
-    global_meta_data.add_to_meta_data("seed", seed)
+    global_meta_data["rng"] = Generator.rng
+    global_meta_data["seed"] = seed
 
 
 def generate_seeds(nseeds):
--- a/bilby/core/utils/random.py
+++ b/bilby/core/utils/random.py
@@ -11,7 +11,10 @@ class Generator:
 
 
 def seed(seed):
+    from .meta_data import global_meta_data
+
     Generator.rng = default_rng(seed)
+    global_meta_data.add_to_meta_data("rng", Generator.rng)
 
 
 def generate_seeds(nseeds):
--- a/bilby/core/utils/random.py
+++ b/bilby/core/utils/random.py
@@ -15,6 +15,7 @@ def seed(seed):
 
     Generator.rng = default_rng(seed)
     global_meta_data.add_to_meta_data("rng", Generator.rng)
+    global_meta_data.add_to_meta_data("seed", seed)
 
 
 def generate_seeds(nseeds):
--- a/bilby/core/sampler/__init__.py
+++ b/bilby/core/sampler/__init__.py
@@ -7,6 +7,7 @@ from ..utils import (
     command_line_args,
     env_package_list,
     get_entry_points,
+    global_meta_data,
     loaded_modules_dict,
     logger,
 )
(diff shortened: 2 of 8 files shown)


bilby/core/utils/random.py
```diff
@@ -11,7 +11,11 @@ class Generator:
 
 
 def seed(seed):
+    from .meta_data import global_meta_data
+
     Generator.rng = default_rng(seed)
+    global_meta_data.add_to_meta_data("rng", Generator.rng)
```
ColmTalbot, line 17: Is using this method instead of just doing `global_meta_data[key] = value` required, or some convention I haven't come across before?

bilby/core/utils/random.py
```diff
@@ -11,7 +11,11 @@ class Generator:
 
 
 def seed(seed):
+    from .meta_data import global_meta_data
```
ColmTalbot, line 14: We don't necessarily have to hide these imports, the meta data will be initialized at import time anyway, right?

ColmTalbot (commented):
This looks like a good start to me.

bilby/core/utils/random.py
mj-will: I thought we may want to add some form of check/validation when adding values, so this would make it easier in future. Happy to revert to something simpler though

bilby/core/utils/random.py
mj-will: I was running into circular imports but that may have been an old thing, I'll have a look

mj-will: Yeah, if I make this a top-level import, then I hit circular import issues. I think this is because everything is imported in the  `__init__.py` files in `bilby`, `bilby.core` and `bilby.core.utils`. We could make an exception for `meta_data` file.

mj-will:
The tests are going to fail until https://github.com/bilby-dev/bilby/pull/867 is merged, since it adds support for properly saving the `cosmology` objects

mj-will:
In what seems to be a reoccurring theme, I realised that adding the `rng` to the result object requires new encoding/decoding functions. For now, I've added them hear but I'm happy to move them to another PR.

bilby/core/utils/random.py
GregoryAshton: You can in principle override the get/set item: https://stackoverflow.com/questions/2390827/how-to-properly-subclass-dict-and-override-getitem-setitem

However, my experience when doing this before is that can introduce unexpected behaviour. But, on the other hand it may be a good idea so that any checks are uniformly applied (e.g. if someone utilises the get/set method instead of using `add_to_metadata`

mj-will: That's a good idea @GregoryAshton. I've used `UserDict` as the base class, since this is less error-prone than inheriting directly from `dict`. This does mean, however, I've had to add encodings for `UserDict` (I add `UserList` as well since they can share the same if statement).

Let me know what you both think

bilby/core/utils/random.py
ColmTalbot: Okay, I'm happy with this, it's a shame we decided to import everything all the time, but it's probably best to stick with for now.

ColmTalbot (approved):
This looks good to me now.

asb5468 (approved):
This looks good to me and the tests cover the new functionality suitably.

PR #875, opened 2024-12-09
TST: add `requires_roqs` mark to tests that use ROQs

mj-will:
I realised that some of the tests that use ROQs aren't marked with the `required_roqs` mark. This means they aren't skipped when running

```bash
pytest --skip-roqs
```

### DIFF
--- a/test/gw/likelihood/marginalization_test.py
+++ b/test/gw/likelihood/marginalization_test.py
@@ -157,6 +157,7 @@ class TestMarginalizedLikelihood(unittest.TestCase):
             bilby.run_sampler(like, new_prior)
 
 
+@pytest.mark.requires_roqs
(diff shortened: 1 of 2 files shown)


test/gw/likelihood/marginalization_test.py
```diff
@@ -157,6 +157,7 @@ def test_run_sampler_flags_if_marginalized_time_is_sampled(self):
             bilby.run_sampler(like, new_prior)
 
 
+@pytest.mark.requires_roqs
```
ColmTalbot, line 160: We should also refactor these tests so they don't all require the ROQ. Happy to defer this to a follow up though.

adivijaykumar (approved):

PR #877, opened 2024-12-10
FEAT: add gwsignal waveform generator

ColmTalbot:
@raffienficiaud this is a quick version of how I had envisioned an extension to the gwsignal support. It doesn't require explicit enumeration of all of the different kinds of source models, users can just do something like

```python
import bilby

priors = bilby.gw.prior.BBHPriorDict()
priors["eccentricity"] = bilby.core.prior.Uniform(0, 1)
priors["mean_per_ano"] = 0.0
del priors["a_1"], priors["a_2"], priors["tilt_1"], priors["tilt_2"], priors["phi_jl"], priors["phi_12"]

wfg = bilby.gw.waveform_generator.GWSignalWaveformGenerator(
    waveform_arguments=dict(waveform_approximant="EccentricFD"),
    duration=4,
    sampling_frequency=1024,
    eccentric=True,
    spinning=False,
)

wfg.frequency_domain_strain(priors.sample())
```

I'm sure there are improvements that could be made to the interface, but I'm hopeful that something like this can be fairly transparent to the underlying `gwsignal` capabilities and ideally not need a lot of modification as more features are added to that. Since you have more experience with `gwsignal` I'd be glad to hear any thoughts.

Also @adivijaykumar @antoniramosbuades and @mj-will

### DIFF
--- a/bilby/gw/waveform_generator.py
+++ b/bilby/gw/waveform_generator.py
@@ -3,6 +3,7 @@ import numpy as np
 from ..core import utils
 from ..core.series import CoupledTimeAndFrequencySeries
 from ..core.utils import PropertyAccessor
+from ..core.utils import logger
 from .conversion import convert_to_lal_binary_black_hole_parameters
 from .utils import lalsim_GetApproximantFromString
 
@@ -63,7 +64,7 @@ class WaveformGenerator(object):
                                                                     start_time=start_time)
         self.frequency_domain_source_model = frequency_domain_source_model
         self.time_domain_source_model = time_domain_source_model
-        self.source_parameter_keys = self.__parameters_from_source_model()
+        self.source_parameter_keys = self._parameters_from_source_model()
         if parameter_conversion is None:
             self.parameter_conversion = convert_to_lal_binary_black_hole_parameters
         else:
@@ -212,7 +213,7 @@ class WaveformGenerator(object):
         dict: The dictionary of parameter key-value pairs
 
         """
-        return self.__parameters
+        return self._parameters
 
     @parameters.setter
     def parameters(self, parameters):
@@ -235,10 +236,10 @@ class WaveformGenerator(object):
         for key in self.source_parameter_keys.symmetric_difference(
                 new_parameters):
             new_parameters.pop(key)
-        self.__parameters = new_parameters
-        self.__parameters.update(self.waveform_arguments)
+        self._parameters = new_parameters
+        self._parameters.update(self.waveform_arguments)
 
-    def __parameters_from_source_model(self):
+    def _parameters_from_source_model(self):
         """
         Infer the named arguments of the source model.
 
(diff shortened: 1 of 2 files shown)


bilby/gw/waveform_generator.py
```diff
@@ -271,3 +272,243 @@ def validate_reference_frequency(self):
...
+            'distance': parameters["luminosity_distance"],
+            'inclination': iota,
+            'eccentricity': parameters["eccentricity"],
+            'longAscNodes': longitude_ascending_nodes,
+            'meanPerAno': parameters["mean_per_ano"],
+            'condition': self.condition
+        }
+        gwsignal_dict = {
+            key: val << Cosmo_units_dictionary.get(key, 0)
+            for key, val in gwsignal_dict.items()
```
mj-will, line 383: What is this bitwise operator doing in this case? I can't quite figure it out.

bilby/gw/waveform_generator.py
```diff
@@ -271,3 +272,243 @@ def validate_reference_frequency(self):
...
+
+        gwsignal_dict.update(extra_args)
+        return gwsignal_dict
+
+    @property
+    def condition(self):
+        condition = 0
+        if self.generator.metadata["implemented_domain"] == 'time':
+            condition = 1
+        return condition
```
mj-will, line 415: Is there a reason to use `0/1` instead of `True/False`?

mj-will:
Thanks for doing things @ColmTalbot, it looks like a good start to me! I had a quick look and added some comments.

bilby/gw/waveform_generator.py
ColmTalbot: It's recommended for performance with astropy quantities (https://docs.astropy.org/en/stable/units/index.html#performance-tips)

bilby/gw/waveform_generator.py
AntoniRamosBuades: The use of 0/1 was added in the first stage of the gwsignal development (see this [notebook](https://git.ligo.org/lscsoft/lalsuite/-/blob/master/lalsimulation/python/lalsimulation/gwsignal/docs/source/examples/testing_conditioning_routines.ipynb) for more details on the conditioning in gwsignal). Unfortunately, I do not know why 0/1 was chosen, instead of  False/True

lorenzopompili00:
Hi @ColmTalbot I also really like this idea! 

Just want to mention that for `bilby_tgr` we also have a `gwsignal_binary_black_hole` function, which is a almost copy of the bilby one with just some extra lines [here](https://git.ligo.org/lscsoft/bilby_tgr/-/blob/master/bilby_tgr/pseob/source.py?ref_type=heads#L347-429). To improve maintainability and reduce code duplication, it would be great to have an implementation that allows the waveform generator for `bilby_tgr` to be a subclass of the `bilby` one, without duplicating too much code as is done right now. I still need to look at this more closely, but I wanted to mention this as something to consider.

raffienficiaud:
Thank you for the PR @ColmTalbot . I am all fine with any changes you make.
I have a questions about this interface: it relies on the fact that we can instanciate the `GWSignalWaveformGenerator` class with parameters: is this compatible with the way WF generators can be instanciated in the configuration of `bilby_pipe`?

raffienficiaud:
> * add this as an option in a future release of bilby_pipe

I guess this is the way to go, but I don't know if we can make this ready by the freeze end of March (according to your email).

> * make the options here something that is passed through the waveform_kwargs

For this, we would need a wrapper class that acts as a "factory" for the implementation suggested in the PR.

I don't know how hard is it to get the change into `bilby_pipe`, and I can give a hand. However, I would like in advance to check with the devs if end of March is something feasible from their perspective.

raffienficiaud:
> I noticed that the previous version was incompatible with parallelisation via multiprocessing due to a pickling issue. The current version will reconstruct the waveform generator every time it is pickled/unpickled. The way we handle multiprocessing, this means that the generator should only be set up once when the pool is created.
> 
> @raffienficiaud do you think this will cause any issues for SEOBNR?

For `pyseobnr`, I do not think there is any issue in pickling. I cannot say anything about the GWsignal part.
This pickling is for multiprocess communication or it is also for disk storage? I am asking because the pickle files have reference to the python path from where the objects are defined, and this is usually an issue.

raffienficiaud:
> I ran some tests and added an example using the new generator. Everything seems to be working well, so I think this is ready for a final round of review.

Excellent, thank you @ColmTalbot

bilby/gw/waveform_generator.py
```diff
@@ -75,29 +76,21 @@ def __init__(self, duration=None, sampling_frequency=None, start_time=0, frequen
...
-        utils.logger.info(
-            "Waveform generator initiated with\n"
-            "  frequency_domain_source_model: {}\n"
-            "  time_domain_source_model: {}\n"
-            "  parameter_conversion: {}"
-            .format(utils.get_function_path(self.frequency_domain_source_model),
-                    utils.get_function_path(self.time_domain_source_model),
-                    utils.get_function_path(self.parameter_conversion))
-        )
+        utils.logger.info(f"Waveform generator instantiated: {self}")
```
mj-will: `logger` is imported but `utils.logger` is used instead, I think this should be consistent.

bilby/gw/waveform_generator.py
```diff
@@ -271,3 +264,288 @@ def validate_reference_frequency(self):
         if SimInspiralGetSpinFreqFromApproximant(waveform_approximant_number) == self.LAL_SIM_INSPIRAL_SPINS_FLOW:
             if self.waveform_arguments["reference_frequency"] != self.waveform_arguments["minimum_frequency"]:
                 raise ValueError(f"For {waveform_approximant}, reference_frequency must equal minimum_frequency")
+
+
+class GWSignalWaveformGenerator(WaveformGenerator):
+    """ A waveform generator with specific checks for GW signals """
```
mj-will: I think this would benefit from a longer doc-string to explain e.g. `spinning` means precessing spins.

bilby/gw/waveform_generator.py
```diff
@@ -271,3 +264,288 @@ def validate_reference_frequency(self):
...
+
+    generator_pickles = False
+
+    def __init__(self, spinning=True, eccentric=False, tidal=False, **kwargs):
+        self.spinning = spinning
```
mj-will, line 320: Is `spinning` the right name for this variable?

examples/gw_examples/injection_examples/eccentric_inspiral.py
```diff
@@ -12,21 +12,20 @@
...
-label = "eccentric_GW150914"
+label = "eccentric"
 bilby.core.utils.setup_logger(outdir=outdir, label=label)
 
-
 injection_parameters = dict(
     mass_1=35.0,
     mass_2=30.0,
     eccentricity=0.1,
+    mean_per_ano=0.0,
```
mj-will, line 29: I'm a bit confused by this change, I thought `EccentricFD` doesn't take a `mean_per_ano` parameter? I see it's fixed below so perhaps we should explain this to avoid confusion.

test/gw/likelihood_test.py
```diff
@@ -35,10 +35,10 @@ def setUp(self):
         self.interferometers.set_strain_data_from_power_spectral_densities(
             sampling_frequency=2048, duration=4
         )
-        self.waveform_generator = bilby.gw.waveform_generator.WaveformGenerator(
+        self.waveform_generator = bilby.gw.waveform_generator.GWSignalWaveformGenerator(
             duration=4,
             sampling_frequency=2048,
-            frequency_domain_source_model=bilby.gw.source.lal_binary_black_hole,
+            waveform_arguments=dict(waveform_approximant="IMRPhenomPv2"),
```
mj-will, line 42: What is the motivation for replacing this rather than having both tests?

test/gw/waveform_generator_test.py
```diff
@@ -695,5 +695,244 @@ def _test_caching_different_domain(func1, func2, params1, params2):
...
+            mean_per_ano=0.0,
+        )
+        parameters_2 = dict(
+            mass_1=10,
+            mass_2=10,
+            theta_jn=1.0,
+            luminosity_distance=1.0,
+            phase=0.0,
+            eccentricity=0.3,
+            mean_per_ano=1.0,
```
mj-will, line 790: I don't think this should do anything with `EccentricFD`

mj-will (changes requested):
This looks good to me. Just a couple of minor comments and questions.

bilby/gw/waveform_generator.py
ColmTalbot:
```suggestion
        logger.info(f"Waveform generator instantiated: {self}")
```

bilby/gw/waveform_generator.py
ColmTalbot: Yeah, a docstring would help I guess, but if this is false it sets all spin parameters to zero (https://github.com/bilby-dev/bilby/pull/877/files#diff-f967bdbab0a8f4f5c8a485178d3fa08d0f1d68def5ba535a1507c7ff0835400aR331). I'll update this docstring also.

test/gw/waveform_generator_test.py
ColmTalbot: That's true, but I think in my tests it does need the parameter. Is there a better choice? I'd rather avoid SEOB just so that it runs faster.

test/gw/likelihood_test.py
ColmTalbot: Mostly just having the new generator tested somewhere in context. This is also a less important test as I don't this likelihood is widely used.

examples/gw_examples/injection_examples/eccentric_inspiral.py
ColmTalbot: Yeah, I'm fairly sure it failed without, the parameter might be needed but not used, I'll check it again later.

ColmTalbot (commented):
Thanks for the comments, I'll address some of the longer ones later.

bilby/gw/waveform_generator.py
mj-will: Ah, okay. I guess we don't distinguish between aligned and precessing spins

bilby/gw/waveform_generator.py
mj-will: @ColmTalbot I think this is all that's left.

mj-will (approved):

bilby/gw/waveform_generator.py
```diff
@@ -284,3 +277,329 @@ def validate_reference_frequency(self):
...
+        from lalsimulation.gwsignal.core import parameter_conventions
+
+        keys = parameter_conventions.nongr_params.copy()
+        if not self.eccentric:
+            keys += ["eccentricity", "mean_per_ano"]
+        if not self.tidal:
+            keys += ["lambda_1", "lambda_2"]
+        if not self.spinning:
+            keys += ["a_1", "a_2", "tilt_1", "tilt_2", "phi_12", "phi_jl"]
+        keys += ["longitude_ascending_nodes"]
```
asb5468: I don't understand what this is doing. What are `nongr_params` and why do we care about them here? Why are we adding eccentricity to the keys if the waveform is not eccentric? Why is `longitude_ascending_nodes` not associated with an if statement?

bilby/gw/waveform_generator.py
```diff
@@ -284,3 +277,329 @@ def validate_reference_frequency(self):
...
+            'inclination': iota,
+            'eccentricity': parameters["eccentricity"],
+            'longAscNodes': parameters["longitude_ascending_nodes"],
+            'meanPerAno': parameters["mean_per_ano"],
+            'condition': self.condition,
+        }
+        gwsignal_dict = {
+            key: val << Cosmo_units_dictionary.get(key, 0)
+            for key, val in gwsignal_dict.items()
+        }
```
asb5468, line 458: What is `Cosmo_units_dictionary` and what is it doing here?

bilby/gw/waveform_generator.py
```diff
@@ -284,3 +277,329 @@ def validate_reference_frequency(self):
...
+
+        gwsignal_dict.update(extra_args)
+        return gwsignal_dict
+
+    @property
+    def condition(self):
+        condition = 0
+        if self.generator.metadata["implemented_domain"] == 'time':
+            condition = 1
+        return condition
```
asb5468: Why is this called `condition` (very vague) and not something like `time_domain`?

examples/gw_examples/injection_examples/eccentric_inspiral.py
```diff
@@ -12,21 +12,20 @@
 import numpy as np
 from bilby.core.utils.random import seed
 
-# Sets seed of bilby's generator "rng" to "123" to ensure reproducibility
```
asb5468, line 15: Why have all the comments been deleted? Since this is an example, not source code, the comments are helpful to understand what each block of code is doing. I would request for them to be added back in.

bilby/gw/waveform_generator.py
ColmTalbot: These were needed at some stage earlier in the implementation as gwsignal was looking for explicit defaults. It looks like they aren't needed now, so I'll remove them.

bilby/gw/waveform_generator.py
ColmTalbot: This sets the appropriate units for each of the parameters. The `<<` operator is a quirk that is recommended by astropy and the specific units are enumerated in gwsignal (and imported further up), I moved the import to here to make it more clear.

bilby/gw/waveform_generator.py
ColmTalbot: Good idea, this name is inherited from gwsignal, but I just removed this and added checks of the implemented domain.

examples/gw_examples/injection_examples/eccentric_inspiral.py
ColmTalbot: These comments tend to get out of date pretty quickly, but that isn't a great reason to remove them. I reinstated the comments, but I updated them a little.

bilby/gw/waveform_generator.py
asb5468: Sorry, can you explain what this block of code is doing? I don't understand how "<<", which Google tells me is a bitwise left shift, applies to units, or what type `Cosmo_units_dictionary` is returning.

bilby/gw/waveform_generator.py
asb5468: Now it seems like `longitude_ascending_nodes` has been entirely removed from this MR? Won't that cause problems? Does a default value need to be set somewhere?

bilby/gw/waveform_generator.py
adivijaykumar: I was similarly confused about this when I saw it, but found https://docs.astropy.org/en/stable/units/quantity.html?utm_source=chatgpt.com#creating-and-converting-quantities-without-copies helpful.

raffienficiaud: I believe it comes from this https://docs.astropy.org/en/latest/units/quantity.html#creating-and-converting-quantities-without-copies

bilby/gw/waveform_generator.py
ColmTalbot: Evidently not, the default was previously being explitictly set to zero, but now it is falling back to whatever the lalsimulation default is.

bilby/gw/waveform_generator.py
mj-will: See my original question here: https://github.com/bilby-dev/bilby/pull/877#discussion_r1880144723

PR #884, opened 2025-01-07
MAINT: avoid call to deprecated `scipy.special.btdtr` and `scipy.special.btdtri` functions

mj-will:
`scipy.special.btdtr` and `scipy.special.btdtri` are deprecated and were removed in `scipy` 1.14.0 (see the docs [here](https://docs.scipy.org/doc/scipy-1.13.0/reference/generated/scipy.special.btdtri.html) and [here](https://docs.scipy.org/doc/scipy-1.13.0/reference/generated/scipy.special.btdtr.html#scipy.special.btdtr)). The PR replaces the import and call with `betainc` and `betaincinv` which are equivalent.

**Motivation**

Since this function has been removed, we have to change the code if we are going to support `scipy` >= 1.14.0.

The CI was failing due to this import. See e.g. https://github.com/bilby-dev/bilby/actions/runs/12650850335

**Testing**

I've run the tests locally and they pass.

**Other notes**

I also reformatted the imports to follow the formatting used by `black`

### DIFF
--- a/bilby/core/prior/analytical.py
+++ b/bilby/core/prior/analytical.py
@@ -1,7 +1,18 @@
 import numpy as np
 from scipy.special import erfinv
-from scipy.special._ufuncs import xlogy, erf, log1p, stdtrit, gammaln, stdtr, \
-    btdtri, betaln, btdtr, gammaincinv, gammainc
+from scipy.special._ufuncs import (
+    xlogy,
+    erf,
+    log1p,
+    stdtrit,
+    gammaln,
+    stdtr,
+    betaln,
+    betaincinv,
+    btdtr,
+    gammaincinv,
+    gammainc,
+)
 
 from .base import Prior
 from ..utils import logger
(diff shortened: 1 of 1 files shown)


mj-will:
I noticed whilst making this change that we're importing functions from `scipy.special._ufuncs` rather than `scipy.special`. Does anyone know why that is?

GregoryAshton (approved):
This looks good to me

GregoryAshton:
> I noticed whilst making this change that we're importing functions from `scipy.special._ufuncs` rather than `scipy.special`. Does anyone know why that is?

I think that the functions may not have existed in scipy.special at first, but only in ufuncs (at least this is my recollection).

mj-will:
Thanks @GregoryAshton and @ColmTalbot, given that I'll change the code to just import from `.special` directly.

mj-will:
> > It seems there may be another change in, I think, `matplotlib` that is breaking a different part of the test suite. I'll have a look and see if I can fix it as a part of this PR as well.
> 
> This is due to an incompatibility between gwpy and matplotlib, that Duncan said will be resolved by a future release of gwpy.

Ah, I see, any thoughts on how to best avoid this? To me, the easiest option would probably be to just set matplotlib <3.10 in the CI.

mj-will:
I'm unsure why the tests are still failing. I've created a local env with matploltib 3.10 and then downgraded to 3.9.4 and the tests then pass. Does anyone have any ideas?

mj-will:
> I'm unsure why the tests are still failing. I've created a local env with matploltib 3.10 and then downgraded to 3.9.4 and the tests then pass. Does anyone have any ideas?

I think I've identified the issue and I'm trying to address it in https://github.com/bilby-dev/bilby/pull/888

adivijaykumar (approved):

PR #886, opened 2025-01-07
Support non-numpy array backends

ColmTalbot:
I've been working on this PR on and off for a few months, it isn't ready yet, but I wanted to share it in case other people had early opinions.

The goal is to make it easier to interface with models/samplers implemented in e.g., JAX, that support GPU/TPU acceleration and JIT compilation.

The general guiding principles are:

- when possible maintain existing behaviour with numpy/builtin arguments
- work introspectively so users don't need to specify the target backend, but use input types
- write as little backend specific code as possible, mostly through using the [`array-api` specification](https://data-apis.org/array-api/latest/index.html) and [`scipy` interoperability](https://docs.scipy.org/doc/scipy/dev/api-dev/array_api.html)

The primary changes so far are:

- making most priors backend independent, there are a few holdouts where the underlying scipy functionality isn't compatible yet
- core likelihoods mostly work with data from any backend
- GW likelihoods work with any backend supported by the source function
- the GW detector objects don't work via introspection, they need to be manually set
- GW geometry (currently in bilby_cython) is handled via multiple-dispatch and added back into bilby

Changed behaviour:

- prior sampling/rescale shapes - related to https://github.com/bilby-dev/bilby/pull/863
- some priors won't return floats on float input

Remaining issues:

- [ ] Saving/loading nun-numpy arrays in result files may not work
- [x] I added some additional parameter conversions that I will remove
- [x] the `bilby.gw.jaxstuff` file should be removed and relevant functionality be moved elsewhere, it's currently just used for testing
- [x] the ROQ likelihood hasn't been ported
- [x] add more testing with JAX
- [x] translate some of the hyperparameter functionality, c.f., [GWPopulation](https://github.com/colmtalbot/gwpopulation)


bilby/compat/patches.py
```diff
@@ -0,0 +1,37 @@
...
+    if aac.is_numpy_namespace(xp):
+        from scipy.special import erfinv
+    elif aac.is_jax_namespace(xp):
+        from jax.scipy.special import erfinv
+    elif aac.is_torch_namespace(xp):
+        from torch.special import erfinv
+    elif aac.is_cupy_namespace(xp):
+        from cupyx.scipy.special import erfinv
+    else:
+        raise BackendNotImplementedError
```
mj-will: I think it would be useful to include the backend in the error.

bilby/compat/utils.py
```diff
@@ -0,0 +1,60 @@
...
+
+import numpy as np
+from array_api_compat import array_namespace
+
+from ..core.utils.log import logger
+
+__all__ = ["array_module", "promote_to_array"]
+
+
+def array_module(arr):
```
mj-will, line 35: This would benefit from a doc-string

bilby/compat/utils.py
```diff
@@ -0,0 +1,60 @@
...
+        elif arr.__module__.startswith("pandas"):
+            return np
+        else:
+            logger.warning(
+                f"Unknown array module for type: {type(arr)} Defaulting to numpy."
+            )
+            return np
+
+
+def promote_to_array(args, backend, skip=None):
```
mj-will: Have you thought about how devices would be handled here?

Moving arrays to a from GPUs can sometimes require more than just calling `array`.

mj-will: Suggest adding a doc-string

bilby/core/prior/analytical.py
```diff
@@ -1,21 +1,25 @@
+import os
+
 import numpy as np
+os.environ["SCIPY_ARRAY_API"] = "1"  # noqa  # flag for scipy backend switching
```
mj-will: I worry slightly about having this hard coded. Does it introduce more overhead when using just numpy?

bilby/core/prior/analytical.py
```diff
@@ -67,10 +71,10 @@ def prob(self, val):
 
         """
         at_peak = (val == self.peak)
-        return np.nan_to_num(np.multiply(at_peak, np.inf))
+        return at_peak * 1.0
```
mj-will, line 72: Is `*1.0` here being used to convert from bool to float?

bilby/core/prior/analytical.py
```diff
@@ -425,15 +419,17 @@ def prob(self, val):
...
+    def cdf(self, val, *, xp=np):
         _cdf = (
-            (np.sin(val) - np.sin(self.minimum))
-            / (np.sin(self.maximum) - np.sin(self.minimum))
-            * (val >= self.minimum) * (val <= self.maximum)
-            + 1 * (val > self.maximum)
+            (xp.sin(val) - xp.sin(self.minimum)) /
+            (xp.sin(self.maximum) - xp.sin(self.minimum))
         )
+        _cdf *= val >= self.minimum
```
mj-will, line 444: I'm not sure the in place operation will work with jax.

bilby/core/prior/analytical.py
```diff
@@ -481,15 +479,17 @@ def prob(self, val):
...
+    def cdf(self, val, *, xp=np):
         _cdf = (
-            (np.cos(val) - np.cos(self.minimum))
-            / (np.cos(self.maximum) - np.cos(self.minimum))
-            * (val >= self.minimum) * (val <= self.maximum)
-            + 1 * (val > self.maximum)
+            (xp.cos(val) - xp.cos(self.minimum))
+            / (xp.cos(self.maximum) - xp.cos(self.minimum))
         )
+        _cdf *= val >= self.minimum
```
mj-will, line 504: More in place operations.

bilby/core/prior/analytical.py
```diff
@@ -627,17 +641,15 @@ def prob(self, val):
...
-        _cdf = (
-            (
-                erf((val - self.mu) / 2 ** 0.5 / self.sigma)
-                - erf((self.minimum - self.mu) / 2 ** 0.5 / self.sigma)
-            ) / 2 / self.normalisation * (val >= self.minimum) * (val <= self.maximum)
-            + 1 * (val > self.maximum)
-        )
+        _cdf = (erf((val - self.mu) / 2 ** 0.5 / self.sigma) - erf(
+            (self.minimum - self.mu) / 2 ** 0.5 / self.sigma)) / 2 / self.normalisation
+        _cdf *= val >= self.minimum
```
mj-will, line 654: In place operation.

bilby/core/prior/analytical.py
```diff
@@ -1012,37 +1004,15 @@ def ln_prob(self, val):
...
-            if np.isfinite(_ln_prob) and self.minimum <= val <= self.maximum:
-                return _ln_prob
-            return -np.inf
-        else:
-            _ln_prob_sub = np.full_like(val, -np.inf)
-            idx = np.isfinite(_ln_prob) & (val >= self.minimum) & (val <= self.maximum)
-            _ln_prob_sub[idx] = _ln_prob[idx]
-            return _ln_prob_sub
+        ln_prob = xlog1py(self.beta - 1.0, -val) + xlogy(self.alpha - 1.0, val)
+        ln_prob -= betaln(self.alpha, self.beta)
```
mj-will: In place operation

bilby/core/prior/analytical.py
```diff
@@ -1235,15 +1196,17 @@ def __init__(self, k, theta=1., name=None, latex_label=None, unit=None, boundary
...
-    def rescale(self, val):
+    @xp_wrap
+    def rescale(self, val, *, xp=np):
         """
         'Rescale' a sample from the unit line element to the appropriate Gamma prior.
 
         This maps to the inverse CDF. This has been analytically solved for this case.
         """
-        return gammaincinv(self.k, val) * self.theta
+        return xp.asarray(gammaincinv(self.k, val)) * self.theta
```
mj-will, line 1226: Does this mean this is falling back to numpy?

bilby/core/prior/analytical.py
```diff
@@ -1789,3 +1750,5 @@ def cdf(self, val):
                 / (self.mode - self.rescaled_minimum)
             )
         )
+
+    betaln,
```
mj-will: Not sure what this is.

bilby/core/prior/dict.py
```diff
@@ -832,13 +842,15 @@ def ln_prob(self, sample, axis=None, normalized=True):
...
             for key in sample
-        ]
-        ln_prob = np.sum(res, axis=axis)
-        return self.check_ln_prob(sample, ln_prob,
-                                  normalized=normalized)
+        ])
+        ln_prob = xp.sum(res, axis=axis)
+        return ln_prob
+        # return self.check_ln_prob(sample, ln_prob,
+        #                           normalized=normalized)
```
mj-will: Is the removal of this intentional?

bilby/core/prior/dict.py
```diff
@@ -862,48 +874,28 @@ def rescale(self, keys, theta):
...
-                result[key] = value
-
-        def safe_flatten(value):
-            """
-            this is gross but can be removed whenever we switch to returning
-            arrays, flatten converts 0-d arrays to 1-d so has to be special
-            cased
-            """
-            if isinstance(value, (float, int)):
-                return value
```
mj-will, line 902: Is removing this intentional?

bilby/gw/utils.py
```diff
@@ -270,6 +247,8 @@ def zenith_azimuth_to_ra_dec(zenith, azimuth, geocent_time, ifos):
     ra, dec: float
         The zenith and azimuthal angles in the sky frame.
     """
+    # delta_x = ifos[0].geometry.vertex - ifos[1].geometry.vertex
+    # theta, phi = zenith_azimuth_to_theta_phi(zenith, azimuth, delta_x)
```
mj-will: Suggest we remove this.

mj-will (commented):
Some initial comments but I'll need to have another look.

bilby/core/prior/analytical.py
ColmTalbot: Yeah, I should update/recheck this, but at least jax doesn't have good support for this, but it looks like tensorflow has a version that numpyro uses (https://github.com/jax-ml/jax/issues/5350). cupy does have this function, so this workaround may have just been for jax. I could add a BackendNotImplementedError.

bilby/core/prior/analytical.py
ColmTalbot: Not anything good.

```suggestion
```

bilby/core/prior/analytical.py
ColmTalbot: Yeah

bilby/core/prior/analytical.py
ColmTalbot: This kind of in-place operation works, it's just operations on slices that don't work. Things like patterns that sometimes existed

```
_cdf = ...
_cdf[val < self.minimum] = 0
_cdf[val > self.maximum] = 1
```

bilby/core/prior/dict.py
ColmTalbot: I'm fairly sure it was, but I'll double check. I think `check_ln_prob` was problematic in some way.

bilby/core/prior/dict.py
ColmTalbot: Yeah, this is in line with one of the other open PRs to update this logic. I'll dig it out in my next pass.

bilby/gw/utils.py
ColmTalbot:
```suggestion
```

bilby/gw/utils.py
```diff
@@ -1038,7 +1017,8 @@ def ln_i0(value):
     array-like:
         The natural logarithm of the bessel function
     """
-    return np.log(i0e(value)) + np.abs(value)
+    xp = array_module(value)
```
ColmTalbot: Comment to self: use `xp_wrap` here.

ColmTalbot (commented):
Thanks for the initial comments @mj-will I'll take a pass at them ASAP.

bilby/compat/patches.py
```diff
@@ -0,0 +1,50 @@
+import array_api_compat as aac
```
GregoryAshton, line 1: Is the idea that this file provides compatibility between all the different array types? Dare I say it, but it feels like this should be a whole python package in itself...

bilby/compat/patches.py
```diff
@@ -0,0 +1,50 @@
...
+
+from .utils import BackendNotImplementedError
+
+
+def erfinv_import(xp):
```
GregoryAshton: All of these functions would benefit from a docstring to explain they do the import given the type of array backend.

bilby/core/prior/analytical.py
GregoryAshton: May be wise to add a comment here and in the other instance in case someone "cleans it up" down the road.

bilby/core/prior/analytical.py
```diff
@@ -720,20 +745,10 @@ def prob(self, val):
...
-            else:
-                _prob = np.exp(-(np.log(val) - self.mu) ** 2 / self.sigma ** 2 / 2)\
-                    / np.sqrt(2 * np.pi) / val / self.sigma
-        else:
-            _prob = np.zeros(val.size)
-            idx = (val > self.minimum)
-            _prob[idx] = np.exp(-(np.log(val[idx]) - self.mu) ** 2 / self.sigma ** 2 / 2)\
-                / np.sqrt(2 * np.pi) / val[idx] / self.sigma
-        return _prob
+        return xp.exp(self.ln_prob(val, xp=xp))
```
GregoryAshton, line 741: I presume there was some reason we handled things in a complicated way before.. was it just bad/lazy coding or where we fixing some edge case? Anyone recall..

bilby/core/prior/analytical.py
```diff
@@ -835,26 +834,13 @@ def ln_prob(self, val):
...
-            if val < self.minimum:
-                _cdf = 0.
-            else:
-                _cdf = 1. - np.exp(-val / self.mu)
-        else:
-            _cdf = np.zeros(val.size)
-            _cdf[val >= self.minimum] = 1. - np.exp(-val[val >= self.minimum] / self.mu)
-        return _cdf
+        with np.errstate(divide="ignore"):
+            return -val / self.mu - xp.log(xp.asarray(self.mu)) + xp.log(val >= self.minimum)
```
GregoryAshton, line 831: Ah okay - are the bounds being implemented here? But, I don't see the upper bound being implemented.

bilby/core/likelihood.py
```diff
@@ -282,9 +285,10 @@ def __init__(self, x, y, func, sigma=None, **kwargs):
 
     def log_likelihood(self, parameters=None):
         parameters = _fallback_to_parameters(self, parameters)
+        xp = array_module(self.x)
         sigma = parameters.get("sigma", self.sigma)
-        log_l = np.sum(- (self.residual(parameters) / sigma)**2 / 2 -
-                       np.log(2 * np.pi * sigma**2) / 2)
+        log_l = xp.sum(- (self.residual(parameters) / sigma)**2 / 2 -
+                       xp.log(xp.asarray(2 * np.pi * sigma**2)) / 2)
```
GregoryAshton, line 204: I'm seeing this in a lot of places. It isn't required by `numpy`. Presumably it is required by `jax` or something else?

bilby/gw/detector/interferometer.py
```diff
@@ -312,11 +325,14 @@ def get_detector_response(self, waveform_polarizations, parameters, frequencies=
         used to set the time at which the antenna response is evaluated,
         otherwise the provided :code:`Parameters["geocent_time"]` is used.
         """
+        xp = array_module(waveform_polarizations)
         if frequencies is None:
-            frequencies = self.frequency_array[self.frequency_mask]
+            # frequencies = self.frequency_array[self.frequency_mask]
```
GregoryAshton: Is there a reason to leave this commented out?

bilby/gw/detector/interferometer.py
```diff
@@ -331,8 +347,8 @@ def get_detector_response(self, waveform_polarizations, parameters, frequencies=
                 antenna_time,
                 parameters['psi'], mode)
 
-            signal[mode] = waveform_polarizations[mode] * det_response
-        signal_ifo = sum(signal.values()) * mask
+            signal[mode] = waveform_polarizations[mode] * mask * det_response
```
GregoryAshton, line 349: It looks like this is changing the way the mask is being used. From operating on a view to operating on the full array but zeroing the False cases. Is that correct?

GregoryAshton (commented):
Okay, I got through about 60% of the diff and I'm pausing here so will submit the questions so far.

bilby/core/prior/analytical.py
ColmTalbot: I think this is carried over from the existing implementation.

bilby/core/likelihood.py
ColmTalbot: Yeah, especially when using jit, jax is pickier than numpy, torch even more so.

bilby/gw/detector/interferometer.py
ColmTalbot: Nope!

```suggestion
```

bilby/gw/detector/interferometer.py
ColmTalbot: I think it isn't, and I've just moved this multiplication by the mask up by a line.
I'm not sure why, but I don't think it should make a big difference.

bilby/compat/patches.py
ColmTalbot: In effect that's what `array_api_compat` does. There are going to be unsupported bits of functionality though. I suspect that over time scipy functions will asymptote towards complete compatibility, but for now, we'll need to have some patches.

Already, in the months since these comments, scipy have added support for erfinv, so one of these patches is being dropped now.

bilby/compat/patches.py
ColmTalbot: Done (for the one remaining function)

bilby/compat/utils.py
ColmTalbot: Done.

bilby/compat/utils.py
ColmTalbot: No, we should probably add this down the line. For now I'm just going to add a comment that this isn't supported.

bilby/core/prior/analytical.py
ColmTalbot: I agree. When we get close to merging I'll take this out.

bilby/core/prior/analytical.py
ColmTalbot: Done.

bilby/core/prior/analytical.py
ColmTalbot: I think it's to avoid division/invalid value in log messages. Those are now explicitly escaped using `numpy.errstate`.

mj-will:
> Python 3.10 doesn't have support for a vmappable version of logsumexp through scipy leading to this job failing (https://github.com/bilby-dev/bilby/actions/runs/25883935510/job/76070707573?pr=886).
> 
> How do people feel about dropping support for Python 3.10 in Bilby 3.10? Numpy dropped support about a year ago.

Sounds reasonable to me. I think we should be looking forward with these changes as much as possible, rather than adding patches/workarounds to support older versions.

bilby/compat/patches.py
mj-will: Just bumping this since it seems to get lots in GitHub hiding old commits

bilby/core/prior/analytical.py
mj-will: Would this be a candidate for a small patch that uses the TF version for jax until jax supports it natively?

bilby/core/prior/analytical.py
GregoryAshton: Does this now need to be taken out? (Reading the mattermost it seems the priority is to get this merged)

bilby/gw/utils.py
GregoryAshton: Does this need to be actioned?

bilby/compat/patches.py
GregoryAshton: Okay

bilby/gw/utils.py
ColmTalbot: This has been done

bilby/core/prior/dict.py
ColmTalbot: Looks like this is back in.

mj-will:
This looks great @ColmTalbot. i've not a proper look yet but one thought on using `orng`: I think `ArrayRNG` may be confusing so I'm considering changing it to something else (see https://github.com/sequince-dev/orng/pull/9). Feel free to comment on the MR if you have thoughts.

Once that's in, I can make a stable release and get in on conda if that would be useful.

.github/workflows/basic-install.yml
```diff
@@ -39,6 +38,8 @@ jobs:
       - name: Test imports
         run: bash test/ci_test_imports.sh
       - name: Test entry points
+        run: |
+      - if: matrix.os != 'windows-latest'
```
mj-will: I think this might be malformed:


```suggestion
        if: matrix.os != 'windows-latest'
```

I am also confused by why this no longer works on windows

ColmTalbot: This stage never worked on windows, previously the whole workflow had been disabled, now the rest of the steps happen. This doesn't work because you can't use bash commands in the same way.

bilby/core/prior/analytical.py
```diff
@@ -325,21 +335,14 @@ def rescale(self, val):
...
-            rescaled = np.empty_like(val)
-            rescaled[vals_less_than_5] = -self.maximum * np.exp(-2 * val[vals_less_than_5] *
-                                                                np.log(self.maximum / self.minimum))
-            rescaled[~vals_less_than_5] = self.minimum * np.exp(np.log(self.maximum / self.minimum) *
-                                                                (2 * val[~vals_less_than_5] - 1))
-            return rescaled
-
-    def prob(self, val):
+        return (
+            xp.sign(2 * val - 1)
```
mj-will, line 345: I think the behaviour at `val=0.5` has changed. Previously it would return `self.minimum` where as now it will be zero since `np.sign(0) = 0`. I'm not sure, but I don't think this is right for a log-uniform.

bilby/core/prior/analytical.py
```diff
@@ -366,19 +370,12 @@ def ln_prob(self, val):
...
-            + (0.5 + norm * np.log(abs(val) / self.minimum))
-            * (val >= self.minimum) * (val <= self.maximum)
-            + 0.5 * (val > -self.minimum) * (val < self.minimum)
-            + 1 * (val > self.maximum)
-        )
-        return _cdf
+    @xp_wrap
+    def cdf(self, val, *, xp=None):
+        asymmetric = xp.log(xp.abs(val) / self.minimum) / xp.log(xp.asarray(self.maximum / self.minimum))
+        return xp.clip(0.5 * (1 + xp.sign(val) * asymmetric), 0, 1)
```
mj-will, line 387: I think this has a similar problem at 0 since `sign(0) = 0` and `asymmetric=-inf`

bilby/core/prior/analytical.py
```diff
@@ -744,30 +743,18 @@ def ln_prob(self, val):
...
+                -(xp.log(xp.maximum(val, xp.asarray(self.minimum))) - self.mu) ** 2 / self.sigma ** 2 / 2
+                - xp.log((2 * np.pi)**0.5 * val * self.sigma)
+            ) + xp.log(val > self.minimum), nan=-xp.inf, neginf=-xp.inf, posinf=-xp.inf)
+
+    @xp_wrap
+    def cdf(self, val, *, xp=None):
+        with np.errstate(divide="ignore"):
+            return 0.5 + erf(
+                (xp.log(xp.maximum(val, xp.asarray(self.minimum))) - self.mu) / self.sigma / np.sqrt(2)
+            ) / 2
```
mj-will, line 766: Does it still return 0 for values below the minimum? I don't think it does but I might be missing something.

bilby/core/prior/analytical.py
```diff
@@ -1012,37 +996,19 @@ def ln_prob(self, val):
...
-                )
-        else:
-            _cdf = np.nan_to_num(betainc(self.alpha, self.beta,
-                                 (val - self.minimum) / (self.maximum - self.minimum)))
-            _cdf[val < self.minimum] = 0.
-            _cdf[val > self.maximum] = 1.
-            return _cdf
+        ln_prob = xlog1py(xp.asarray(self.beta - 1.0), -val) + xlogy(xp.asarray(self.alpha - 1.0), val)
+        ln_prob -= betaln(xp.asarray(self.alpha), xp.asarray(self.beta))
+        return xp.nan_to_num(ln_prob, nan=-xp.inf, neginf=-xp.inf, posinf=-xp.inf)
```
mj-will: I'm sure minimum and maximum are respected here.

bilby/core/prior/analytical.py
```diff
@@ -1074,27 +1040,19 @@ def __init__(self, mu, scale, name=None, latex_label=None, unit=None, boundary=N
...
-                rescaled = self.mu + self.scale * np.log(val / (1. - val))
-        else:
-            rescaled = np.inf * np.ones(val.size)
-            rescaled[val == 0] = -np.inf
-            rescaled[(val > 0) & (val < 1)] = self.mu + self.scale\
-                * np.log(val[(val > 0) & (val < 1)] / (1. - val[(val > 0) & (val < 1)]))
-        return rescaled
+        with np.errstate(divide="ignore"):
+            val = xp.asarray(val)
+            return self.mu + self.scale * xp.log(xp.maximum(val / (1 - val), xp.asarray(0)))
```
mj-will, line 1070: I don' think the upper boundary is consistent with the old implementation. I think it now gives -inf rather than +inf

mj-will (commented):
I'm unsure if all the priors are consistent with the previous versions. I've tried to go through them but may have missed some things or equally, got some logic wrong and they're fine.

bilby/core/utils/random.py
```diff
@@ -104,3 +108,82 @@ def generate_seeds(nseeds):
...
+    elif isinstance(random_state, np.random.Generator):
+        return np
+    elif aac.is_jax_array(random_state) or getattr(random_state, "backend") == "jax":
+        import jax.numpy as jnp
+        return jnp
+    elif aac.is_torch_array(random_state) or getattr(random_state, "backend") == "torch":
+        import torch
+        return torch
+    else:
+        return np
```
mj-will, line 189: Should this be using the wrapped modules, e.g. `array_api_compat.numpy`?

From [the docs](https://data-apis.org/array-api-compat/helper-functions.html#array_api_compat.array_namespace) I think this is preferred:

> Wrapped array namespaces can also be imported directly. For example, array_namespace(np.array(...)) will return array_api_compat.numpy. This function will also work for any array library not wrapped by array-api-compat if it explicitly defines [__array_namespace__](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__array_namespace__.html) (the wrapped namespace is always preferred if it exists).

bilby/core/utils/series.py
```diff
@@ -150,44 +158,65 @@ def _check_legal_sampling_frequency_and_duration(sampling_frequency, duration):
...
+
     white_noise = re1 + 1j * im1
 
     # set DC and Nyquist = 0
-    white_noise[0] = 0
-    # no Nyquist frequency when N=odd
-    if np.mod(number_of_samples, 2) == 0:
-        white_noise[-1] = 0
+    white_noise = xpx.at(white_noise, 0).set(0)
+    white_noise = xpx.at(white_noise, -1).set(0)
```
mj-will, line 192: May be missing something, but it seems the case where the number of samples is odd is no longer handled.

bilby/core/grid.py
```diff
@@ -57,8 +63,16 @@ def __init__(self, likelihood=None, priors=None, grid_size=101,
...
         gzip: bool
             Set whether to gzip the output grid file
+        xp: array module | None
+            The array module to use for calculations (e.g., :code:`numpy`,
+            :code:`cupy`). If :code:`None`, defaults to :code:`numpy`.
+
         """
 
+        if xp is None:
+            xp = np
```
mj-will, line 72: Should this default to `aac.numpy`?

bilby/core/grid.py
```diff
@@ -150,7 +169,7 @@ def marginalize(self, log_array, parameters=None, not_parameters=None):
         else:
             raise TypeError("Parameters names must be a list or string")
 
-        out_array = log_array.copy()
+        out_array = copy(log_array)
```
mj-will: I'm not sure a plain copy works for torch:

```
>>> import torch
>>> import copy
>>> x = torch.Tensor([1.0])
>>> y = copy.copy(x)
>>> y[0] = 0
>>> x
tensor([0.])
>>> 
```

I'm not sure what the correct way is to handle this, in other packages I've resorted to helpers but I suspect there's a better way. For example: https://github.com/mj-will/aspire/blob/main/src/aspire/utils.py#L456

mj-will: I'm not sure a plain copy works for torch:

```
>>> import torch
>>> import copy
>>> x = torch.Tensor([1.0])
>>> y = copy.copy(x)
>>> y[0] = 0
>>> x
tensor([0.])
>>> 
```

I'm not sure what the correct way is to handle this, in other packages I've resorted to helpers but I suspect there's a better way. For example: https://github.com/mj-will/aspire/blob/main/src/aspire/utils.py#L456

bilby/core/prior/analytical.py
ColmTalbot: This hadn't occurred to me, but I actually think the old version is more incorrect. I'm happy to reinstate the old behaviour with `xp.where`, but it is worth documenting as I think this is non intuitive.

bilby/core/prior/analytical.py
ColmTalbot: Yeah, `max(val, minimum)` is equivalent to a one-sided clip. This seems overly complicated though as minimum is hardcoded to zero, but that's a carry over.

bilby/core/prior/analytical.py
ColmTalbot: Yeah, this seems pretty broken. What about this?


```suggestion
        normalized = (val - self.minimum) / (self.maximum - self.minimum)
        ln_prob = xlog1py(xp.asarray(self.beta - 1.0), -normalized) + xlogy(xp.asarray(self.alpha - 1.0), normalized)
        ln_prob -= betaln(xp.asarray(self.alpha), xp.asarray(self.beta))
        ln_prob -= xlogy(
            xp.asarray(self.alpha + self.beta - 1),
            xp.asarray(self.maximum - self.minimum),
        )
        return xp.where(
            xp.abs(normalized - 0.5) <= 0.5,
            xp.nan_to_num(ln_prob, nan=-xp.inf, neginf=-xp.inf, posinf=-xp.inf),
            -xp.inf,
        )
```

bilby/core/prior/analytical.py
ColmTalbot: I don't think so, `log(max(1 / 0, 0)) = log(inf) = inf`.

bilby/core/prior/analytical.py
ColmTalbot: Yeah, I think we need to modify the definition of `asymmetric` to the following to make the CDF flat between [-min, min].

```python
        asymmetric = xp.log(xp.maximum(xp.abs(val) / self.minimum, 1)) / xp.log(xp.asarray(self.maximum / self.minimum))
```

bilby/core/utils/series.py
ColmTalbot: I think the answer is that the old function was broken... and always returned yes in this condition.
We could add it back in, but it does break the specific value tests we have.

bilby/core/grid.py
ColmTalbot: Sigh

bilby/core/utils/random.py
ColmTalbot: I'm not against doing this, but I'm not really sure what the benefit is. I guess it would reduce the number of explicit imports.

ColmTalbot (commented):
Thanks for the comments @mj-will I'll try to fold in the changes before next week.

bilby/core/utils/random.py
mj-will: I think the main benefit is the APIs are standardized and the code can therefore be more generic.

For example, the methods will support some additional keywords like `device`:
```
import array_api_compat.numpy as np
device = "cpu"
x = [1, 2, 3]
np.asarray(x, device=device)
```
This doesn't do anything for numpy in this case, but doesn't raise an error like it would for standard numpy. This means the exact same code would work for torch with e.g. a CUDA device.

bilby/core/prior/analytical.py
mj-will: Ah, I was thinking specifically of things above the bound, so:

```
log(max(negative, 0)) = log(0) = -inf
```

I agree the bound itself is still the same.

bilby/core/utils/random.py
ColmTalbot: Okay, I'll try to use this where I can, but I think we should avoid using the array_api_compat version of numpy if `BILBY_ARRAY_API == False`.

FWIW, this specific example has worked with standard numpy since version 2.0.0.

bilby/core/prior/analytical.py
ColmTalbot: Oh, I don't know that we've been careful about out of bounds behaviour for these rescale methods in general. The behaviour here is the same as the behaviour before this change for scalars, but not arrays.

bilby/core/prior/analytical.py
mj-will: Happy to stick with the new version, but I think we should note that it has changed in the doc-string.

mj-will:
> > i've not a proper look yet but one thought on using orng: I think ArrayRNG may be confusing so I'm considering changing it to something else (see [sequince-dev/orng#9](https://github.com/sequince-dev/orng/pull/9)). Feel free to comment on the MR if you have thoughts.
> 
> No meaningful thoughts, I agree it's probably a positive change and I'm happy to update once it is available.
> 
> > Once that's in, I can make a stable release and get in on conda if that would be useful.
> 
> I think we'll definitely want this once we have a release ready. Although, I think I've refactored things so that orng doesn't need to be a top-level dependency.

[v0.1.0a6](https://github.com/sequince-dev/orng/releases/tag/v0.1.0a6) and above now renames `ArrayRNG` to `RandomGenerator`

docs/array_api.rst
```diff
@@ -0,0 +1,552 @@
...
+the operation using the ``random_state`` parameter,
+as there's no input array to infer the backend from:
+
+.. code-block:: python
+
+    import bilby
+    import jax
+    
+    prior = bilby.core.prior.Uniform(minimum=0, maximum=10)
+    samples = prior.sample(size=1000, random_state=jax.random.PRNGKey(42))  # Returns JAX array
```
mj-will: I think `PRNGKey` is now deprecated in favour of `jax.random.key(42)`: https://docs.jax.dev/en/latest/jax.random.html#prng-keys

mj-will:
@ColmTalbot orng 0.1.0 is now on conda forge: https://anaconda.org/channels/conda-forge/packages/orng/overview

bilby/core/grid.py
ColmTalbot: See https://github.com/bilby-dev/bilby/pull/886#discussion_r3288131657

bilby/core/prior/analytical.py
ColmTalbot: This isn't natively supported by either JAX or torch, but will be evaluated on the CPU by scipy and then returned as the relevant array type. Personally, I would just leave this as is. I can add a comment to the docstring that this is not well supported.

bilby/core/prior/analytical.py
ColmTalbot: I'm going to say this is not in scope for this change as it preserves the existing behaviour and we should probably have a more generic answer for what happens when people pass values outside [0, 1] to the rescale methods.

bilby/core/prior/analytical.py
mj-will: Reminder to me: check this

GregoryAshton (approved):
I'm happy to approve and merge this with the idea to then ask all developers to test it and find the issues.

bilby/core/prior/analytical.py
```diff
@@ -1012,37 +1005,28 @@ def ln_prob(self, val):
         =======
         Union[float, array_like]: Prior probability of val
         """
-        _ln_prob = xlogy(self.alpha - 1, val - self.minimum) + xlogy(self.beta - 1, self.maximum - val)\
-            - betaln(self.alpha, self.beta) - xlogy(self.alpha + self.beta - 1, self.maximum - self.minimum)
+        normalized = (val - self.minimum) / (self.maximum - self.minimum)
+        ln_prob = (
+            xlog1py(xp.asarray(self.beta - 1.0), -normalized)
+            + xlogy(xp.asarray(self.alpha - 1.0), normalized)
+            - xlogy(xp.asarray(self.alpha + self.beta - 1), xp.asarray(self.maximum - self.minimum))
```
mj-will: I don't think this is right anymore. We're using the normalized space so it should be the PDF there plus the jacobian term

Instead, I think we need the Jacobian from transforming to the log-space so:


```suggestion
            -  xp.log(self.maximum - self.minimum)
```

I might be missing something though

ColmTalbot: I think this matches https://preliz.readthedocs.io/en/latest/distributions/gallery/beta_scaled.html

ColmTalbot: Oh, I see now.

bilby/core/prior/analytical.py
mj-will: I think this still has an issue between [-1, 1] but I think that was present in the previous version as well.

I've opened an issue to track these sorts of things: https://github.com/bilby-dev/bilby/issues/1102

mj-will (approved):
Thanks again Colm for all your work on this!

Assuming the CI passes, I'm happy to approve this and we can continue to test it.

mj-will (approved):

mj-will (approved):

PR #888, opened 2025-01-08
CI: specify python version and activate env

mj-will:
Run unit tests of different python versions and activate conda environment.

**Motivation**

While try to fix the CI issues in https://github.com/bilby-dev/bilby/pull/884, I realised that the conda environment created in the containers is not being activated. This mean that an `conda` commands are not being run in the right env. This can be fixed by activating the environment. Since the environments are named `python<version without dot>` e.g. `python311`, I thought this would a good moment to add different Python versions into the CI.

**Changes**

I've added a `matrix` entry to the `stratergy` section of the unit test CI. This contains the different Python versions along with the short version name (e.g. 311) which is needed to fetch the container.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -16,6 +16,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  CONDA_PATH: /opt/conda/
+
 jobs:
   build:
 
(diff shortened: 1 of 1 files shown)


mj-will:
It seems the pre-commit checks are now failing for unrelated reasons. I suggest we fix that in a different PR.

GregoryAshton (approved):
This looks good - the matrix is a nice feature!

ColmTalbot (approved):
We'll need to update the branch protections after this to make sure that the correct unit tests are required, but that is a repo settings thing so we can merge this.

mj-will:
Thanks both, I'm going to bypass the branch protections since they're failing for unrelated reasons that should be fixed in other PRs

mj-will:
> We'll need to update the branch protections after this to make sure that the correct unit tests are required, but that is a repo settings thing so we can merge this.

Now that this is merged, I've added updated the branch protection rules to include the 3 unit test jobs

PR #890, opened 2025-01-08
BUG: Update the resume file check to see if the file is empty

### LINKED ISSUE #43: Confusing message on reading empty resume files
The error message presented [here](https://github.com/bilby-dev/bilby/blob/497b7b88a85218814e7bb6f8463fe999a06e495a/bilby/core/sampler/dynesty.py#L758) on starting dynesty sampling can be confusing. Specifically, in `bilby_pipe`, an empty file is created and transferred at the initialisation (I forget why). Therefore, there is an initial message that the result file is corrupted, which could be confusing to new members.

I'd suggest as an easy fix we test if the file is empty and add a statement that the resume file is empty and is hence ignored.

GregoryAshton:
Resolves #43 by adding a check to see if the file is empty

### DIFF
--- a/bilby/core/sampler/dynesty.py
+++ b/bilby/core/sampler/dynesty.py
@@ -743,7 +743,9 @@ class Dynesty(NestedSampler):
         from ... import __version__ as bilby_version
 
         versions = dict(bilby=bilby_version, dynesty=dynesty_version)
-        if os.path.isfile(self.resume_file):
+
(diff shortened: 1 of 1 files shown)


ColmTalbot (approved):

adivijaykumar (approved):
Since we just need the file size and nothing else, I think we could just use `os.path.getsize(self.resume_file)`, but I think this works as well!

PR #893, opened 2025-01-22
Added WeightedCategorical-Prior

JasperMartins:
This pull request adds a `WeightedCategorical`-prior that subsumes the existing `Categorical`-Prior. Functionally, the only difference between the two is that `WeightedCategorical` supports the setting of relative weights for all categories.


mj-will:
Hi @JasperMartins, thank for implementing this and apologies for the delay in looking over it.

The changes look good to me but following a recent change to the `Categorical` prior (see https://github.com/bilby-dev/bilby/pull/947) I think this will need slightly reworking to address the conflicts. Do you think you'd have time/capacity to do this?

bilby/core/prior/analytical.py
```diff
@@ -1482,8 +1509,31 @@ def rescale(self, val):
...
+        float: cumulative prior probability of val
+        """
+        if (not hasattr(val, "__len__")):
+            if val in self.values:
+                return self._cum_p[np.searchsorted(self.values, val)]
+            else:
+                return 0
+        else:
+            index = np.searchsorted(self.values, val)
+            index[index == self.nvalues] = self.nvalues - 1
```
mj-will: Could you help me understand why this is needed?

I understand that if `searchsorted` is given a value greater than the last entry in `values`, then it will return the last index+1. When will this happen? Since this is discrete I thought the values should always match exactly and that seems to give the right index, e.g.

````python
>>> np.searchsorted([1, 3, 5], 5)
np.int64(2)
```

mj-will (commented):
LGTM, just one question from my side.

bilby/core/prior/analytical.py
JasperMartins: If `val` is not in `self.values` (and greater than the last value), it can happen, and then the next line (`self.values[index]`) would fail in an obtuse way.
A more elegant solution would probably be to use `np.where` to check for equality from the start.
This actually made me wonder, though, whether returning 0 is actually the desired behavior of the CDF if `val` is not strictly in `self.values`.

ColmTalbot: Since this is a CDF, if the value is less (more) than the lowest it should return 0 (1). I think this can instead be

```suggestion
            index = np.clip(index, 0, self.nvalues - 1)
```

bilby/core/prior/analytical.py
```diff
@@ -1439,33 +1439,60 @@ def ln_prob(self, val):
...
+        # check for consistent shape of input
+        if len(p) != self.nvalues or len(p.shape) != 1:
+            raise ValueError(
+                "Inconsistent shape of weights and number of categories:"
+                + f"np.atleast_1d(weights) has shape {p.shape} "
+                + f"while number of categories is {self.values}"
+            )
+        self.p = p[sorter]
+        # save cdf for rescaling
+        self._cum_p = np.cumsum(p)
```
ColmTalbot: Rename to `self._cumulative_p`

JasperMartins: In the same vein, if the value falls between the discrete values, it should also return the CDF of the next-lower discrete value, right? This also simplifies the code, actually.

ColmTalbot: Yeah, I checked and this is consistent with what scipy does.

JasperMartins:
Hi @ColmTalbot,

Sorry for the delay, I had been on vacation.
So, prompted by the init-from-repr issue, I have added the new priors to the general assembly of prior-tests, to catch other issues, which also uncovered a small bug in the cdf function for values below the minimum value in the prior-support.
I have fixed these issues and brought the branch up to date.

In principle, the tests in the file `analytical_test.py` could now probably be dropped.

PR #894, opened 2025-01-23
Align Einstein Telescope geometry with LAL

### LINKED ISSUE #713: Einstein Telescope's geometry in Bilby is different from definitions in LAL
In GitLab by @git.ligo:qian.hu on Nov 11, 2022, 10:52

The "ET1", "ET2", "ET3" defined in Bilby have different geometry (and therefore response functions) from [that defined in LAL](https://lscsoft.docs.ligo.org/lalsuite/lal/_l_a_l_detectors_8h_source.html#l00602). 

Not an urgent issue but it's better to get them aligned at some point.

milanwils:
**Problem**
As mentioned in #713, the current geometry of the ET is not in line with its definition in LAL.

1. The labels of ET2 and ET3 are swapped. 
2. Currently the new corner points are calculated by updating the latitude and longitude approximately and not updating the elevation. Additionally, the azimuth and tilt do not take into account the curvature of the earth. This leads to small differences  in the vertex position and unit vector. As a result, the null stream is not accurate and signal leaks into it.

**Proposed Solution**
1. This is addressed in this PR by adding an optional clockwise option to the detector config. I set the value in the `ET.interferometer` file to `clockwise=False`  so that the default is consistent with LAL. If this is undesired for backwards compatibility, it could be set to `clockwise=True` .
2.  This issue is addressed by following the same specification as LAL: [T1400308](https://dcc.ligo.org/cgi-bin/private/DocDB/ShowDocument?.submit=Identifier&docid=T1400308). The code is rather verbose to avoid confusion but can be changed if denser code is preferred. This algorithm required the implementation of a transform from WGS84 coordinated to ECEF coordinates. I added a small test to check the consistency with the forward transformation. In the implementation of this transform I used uninformative variable names to be consistent with the reference (they do not have a physical meaning anyway).

**Additional information**
There is also a small error in the LAL config, see [LAL PR](https://git.ligo.org/lscsoft/lalsuite/-/merge_requests/2399). Only if both get accepted do the configurations become identical.

### DIFF
--- a/bilby/gw/detector/networks.py
+++ b/bilby/gw/detector/networks.py
@@ -310,6 +310,7 @@ class TriangularInterferometer(InterferometerList):
         yarm_azimuth,
         xarm_tilt=0.0,
         yarm_tilt=0.0,
+        clockwise=True,
     ):
         super(TriangularInterferometer, self).__init__([])
         self.name = name
@@ -322,6 +323,7 @@ class TriangularInterferometer(InterferometerList):
             maximum_frequency = [maximum_frequency] * 3
 
         for ii in range(3):
+
             self.append(
                 Interferometer(
                     "{}{}".format(name, ii + 1),
@@ -339,8 +341,8 @@ class TriangularInterferometer(InterferometerList):
                 )
             )
 
-            xarm_azimuth += 240
-            yarm_azimuth += 240
+            xarm_azimuth += 240 if clockwise else 120
+            yarm_azimuth += 240 if clockwise else 120
 
             latitude += (
                 np.arctan(
--- a/bilby/gw/detector/networks.py
+++ b/bilby/gw/detector/networks.py
@@ -2,11 +2,13 @@ import os
 
 import numpy as np
 import math
+from scipy.spatial.transform import Rotation
 
 from ...core import utils
 from ...core.utils import logger, safe_file_dump
 from .interferometer import Interferometer
 from .psd import PowerSpectralDensity
+from ..utils import get_vertex_position_geocentric, get_vertex_position_ellipsoid
 
 
 class InterferometerList(list):
@@ -341,29 +343,48 @@ class TriangularInterferometer(InterferometerList):
                 )
             )
 
-            xarm_azimuth += 240 if clockwise else 120
-            yarm_azimuth += 240 if clockwise else 120
+            unit_vector_x = self[ii].geometry.unit_vector_along_arm("x")
+            unit_vector_y = self[ii].geometry.unit_vector_along_arm("y")
 
-            latitude += (
-                np.arctan(
-                    length
-                    * np.sin(xarm_azimuth * np.pi / 180)
-                    * 1e3
-                    / utils.radius_of_earth
-                )
-                * 180
-                / np.pi
-            )
-            longitude += (
-                np.arctan(
-                    length
-                    * np.cos(xarm_azimuth * np.pi / 180)
-                    * 1e3
-                    / utils.radius_of_earth
-                )
-                * 180
-                / np.pi
-            )
+            vertex_geocentric = get_vertex_position_geocentric(self[ii].latitude_radians,
+                                                               self[ii].longitude_radians,
+                                                               self[ii].elevation)
+            next_vertex_geocentric = vertex_geocentric + length * 1000 * (unit_vector_y if clockwise else unit_vector_x)
+            next_vertex_ellipsoid = get_vertex_position_ellipsoid(next_vertex_geocentric[0],
+                                                                  next_vertex_geocentric[1],
+                                                                  next_vertex_geocentric[2])
+            next_latitude_rad, next_longitude_rad, next_elevation = next_vertex_ellipsoid
+
+            rotation_vector = np.cross(unit_vector_x, unit_vector_y)
+            rotation_vector /= np.linalg.norm(rotation_vector)
+            rotation_angle = - 2 / 3 * np.pi if clockwise else 2 / 3 * np.pi
+            rotation = Rotation.from_rotvec(rotation_angle * rotation_vector)
+            next_unit_vector_x = rotation.apply(unit_vector_x)
+            next_unit_vector_y = rotation.apply(unit_vector_y)
+
+            next_local_normal_vector = np.array([np.cos(next_latitude_rad) * np.cos(next_longitude_rad),
+                                                 np.cos(next_latitude_rad) * np.sin(next_longitude_rad),
+                                                 np.sin(next_latitude_rad)])
+            next_local_north_vector = np.array([-np.sin(next_latitude_rad) * np.cos(next_longitude_rad),
+                                                -np.sin(next_latitude_rad) * np.sin(next_longitude_rad),
+                                                np.cos(next_latitude_rad)])
+            next_local_east_vector = np.array([-np.sin(next_longitude_rad),
+                                               np.cos(next_longitude_rad), 0])
+
+            next_xarm_tilt_rad = np.arcsin(np.dot(next_unit_vector_x, next_local_normal_vector))
+            next_yarm_tilt_rad = np.arcsin(np.dot(next_unit_vector_y, next_local_normal_vector))
+            next_xarm_azimuth_rad = np.arctan2(np.dot(next_unit_vector_x, next_local_north_vector),
+                                               np.dot(next_unit_vector_x, next_local_east_vector))
+            next_yarm_azimuth_rad = np.arctan2(np.dot(next_unit_vector_y, next_local_north_vector),
+                                               np.dot(next_unit_vector_y, next_local_east_vector))
+
+            latitude = np.rad2deg(next_latitude_rad)
+            longitude = np.rad2deg(next_longitude_rad)
+            elevation = next_elevation
+            xarm_azimuth = np.rad2deg(next_xarm_azimuth_rad)
+            yarm_azimuth = np.rad2deg(next_yarm_azimuth_rad)
+            xarm_tilt = next_xarm_tilt_rad
+            yarm_tilt = next_yarm_tilt_rad
 
 
 def get_empty_interferometer(name):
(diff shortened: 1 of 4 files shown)


ColmTalbot (commented):
This is a nice change, I think we'll need to think a little about the best way to do this. I've suggested waiting a while to get this in as I don't think it is urgent, the labelling of ET2/ET3 doesn't really matter, and we can make the changes more satisfyingly if we break things.

bilby/gw/detector/networks.py
```diff
@@ -339,29 +343,48 @@ def __init__(
...
-                np.arctan(
-                    length
-                    * np.cos(xarm_azimuth * np.pi / 180)
-                    * 1e3
-                    / utils.radius_of_earth
-                )
-                * 180
-                / np.pi
-            )
+            unit_vector_x = self[ii].geometry.unit_vector_along_arm("x")
```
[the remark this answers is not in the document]
milanwils: Perhaps we could add setters to the `InterferometerGeometry` class for `x`, `y` and `vertex`? For the lat, long, elevation, tilt and azimuth these setters are separate but it does not seem sensible to me that someone can change only one Cartesian coordinate.

All the update logic could then be moved into the geometry class and only the calculation of the new corner points (which is triangle specific) remains in the 'TriangularInterferometer` class.

My main concern with this is approach is that one would need to be careful to update the vertex position before changing the unit vector. Otherwise the calculation of tilt and azimuth is wrong. This could be avoided by making the unit vectors the "base information" and deriving the tilt and azimuth from it.

This also raises the question of whether the constructor should be adapted to allow users to define the detector directly from Cartesian parameters. I guess you could argue that this is actually the only way you should be able to set these parameters as interferometers are constant.

bilby/gw/detector/networks.py
```diff
@@ -310,6 +312,7 @@ def __init__(
         yarm_azimuth,
         xarm_tilt=0.0,
         yarm_tilt=0.0,
+        clockwise=True,
```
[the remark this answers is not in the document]
milanwils: Fine by me, if people really want they can still override the detector names

bilby/gw/utils.py
```diff
@@ -86,6 +86,49 @@ def get_vertex_position_geocentric(latitude, longitude, elevation):
     return np.array([x_comp, y_comp, z_comp])
 
 
+def get_vertex_position_ellipsoid(x_comp, y_comp, z_comp):
```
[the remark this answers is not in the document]
milanwils: I went through the exact same thought process when writing the function. This seems like a good compromise.

milanwils:
Update: LAL [MR2399](https://git.ligo.org/lscsoft/lalsuite/-/merge_requests/2399) has been approved and the methodology in [T1400308](https://dcc.ligo.org/cgi-bin/private/DocDB/ShowDocument?.submit=Identifier&docid=T1400308) now contains the algorithm added in this PR.

mj-will:
Now we're preparing for 3.0, I think we can more easily consider the breaking changes mentioned above. @milanwils @ColmTalbot can you revisit this and assess where we stand?

milanwils:
@mj-will Sorry for the delay in replying. I had to refresh my memory a bit because I made this request quite some time ago. As far as I can see, the fix in #1016 solved most of the issues but to be entirely correct, the changes in this PR should still be applied because:
- The distance parameter is used as the distance along the surface of the earth, not the distance of the detector arm
- The elevation is ignored in the calculations
- The azimuth is only updated approximately
- The tilt is not updated (local plane at each corner point is slightly different)

I believe there are just a few questions to be answered and then we can proceed rather quickly:
1. Do we want to make the labelling of the detectors consistent with LAL? I believe this is the only breaking change.
2. Do we remove the clockwise argument I introduced? I suggest yes, given that @ColmTalbot  was in favour.
3. What do we do with the function signatures of e.g. `get_vertex_position_ellipsoid`? For me it is fine to use @ColmTalbot 's suggestion.
4. Do we want to use astropy for the conversion of coordinates or the implementation I made?
5. Where to put the logic for calculating the next detector parameters. Currently this is in the initialiser but this can be put into a member function of the class. This would return 7 parameters, which might be a bit cumbersome. I would personally just prefer a decent docstring with the initialiser function.

mj-will:
Thanks @milanwils, here are my thoughts.

>     1. Do we want to make the labelling of the detectors consistent with LAL? I believe this is the only breaking change.
Yes, I think so.

>     2. Do we remove the clockwise argument I introduced? I suggest yes, given that @ColmTalbot  was in favour.
Agree, this would be a breaking change but that's fine for 3.0

>     3. What do we do with the function signatures of e.g. `get_vertex_position_ellipsoid`? For me it is fine to use @ColmTalbot 's suggestion.
Let's start with that and see

>     4. Do we want to use astropy for the conversion of coordinates or the implementation I made?
My concern with astropy would be speed. Do you know how they compare? If your implementation is faster then a test to show they're equivalent may be enough.

>     5. Where to put the logic for calculating the next detector parameters. Currently this is in the initialiser but this can be put into a member function of the class. This would return 7 parameters, which might be a bit cumbersome. I would personally just prefer a decent docstring with the initialiser function.

I've not looked at the PR in detail, but if I'm following correctly, you're referring to the large block of code that's been added to the the initialiser. My preference would be to try and factor this out into a method (or methods) so it's easier to test in isolation but this might need some more thought.

milanwils:
@mj-will 

I have pushed changes for 1-3.

4. Is speed really important here? This function is only supposed to be called when creating the detector object or when debugging I suppose. Even if it would be a fraction slower, I don't see how that would cause any problem.
5.  Ok, I will try to move this logic somewhere else.

milanwils:
I took a look at the array backend flexibility and it should indeed not be hard to implement. To test that everything is working I have to either rebase onto or merge main into this feature branch because the branches diverted before backend flexibility was added to Bilby. What is the standard on this project?

milanwils:
I went with a rebase because it was quite painful.

I stuck with my own implementation for the coordinate transform, moved the update logic to a member function of the TriangularInterferometer class  (the update is specific to such a design) and I have added some array back-end flexibility code that works to the best of my knowledge. For the operations on scalars, I left in the numpy code.

As far as I can see, all requested changes are now included. Let me know if there are more changes I have to make. I will still verify that LAL and Bilby are now in perfect alignment.

Are there additional tests we want to add, e.g. is the sum of the antenna patterns zero?

milanwils:
[test_et_lal_consistency.py](https://github.com/user-attachments/files/32344773/test_et_lal_consistency.py)
The new implementation is identical to the LAL implementation as expected.

bilby/gw/utils.py
```diff
@@ -88,6 +96,46 @@ def get_vertex_position_geocentric(latitude, longitude, elevation):
...
+        3D representation of the geocentric vertex position in meters (x, y, z)
+
+    Returns
+    =======
+    array_like: A 3D representation of the ellipsoidal vertex position (latitude [rad], longitude [rad], elevation [m])
+    """
+    semi_major_axis = 6378137  # for ellipsoid model of Earth, in m
+    semi_minor_axis = 6356752.314  # in m
+
+    r = np.sqrt(position[0] ** 2 + position[1] ** 2)
```
ColmTalbot, line 121: All of the uses of `np` in this function (and the others) should be `xp`.

test/gw/utils_test.py
```diff
@@ -277,6 +277,17 @@ def test_lalsim_SimInspiralChooseFDWaveform(self):
                 1.5,
             )
 
+    def test_get_vertex_position_geocentric_ellipsoid_conversion(self):
+        latitude_true = np.deg2rad(46 + 27. / 60 + 18.528 / 3600)
+        longitude_true = np.deg2rad(-(119 + 24. / 60 + 27.5657 / 3600))
+        elevation_true = 142.554
+        vertex = gwutils.get_vertex_position_geocentric(
+            self.xp.asarray([latitude_true, longitude_true, elevation_true]))
+        latitude, longitude, elevation = gwutils.get_vertex_position_ellipsoid(vertex)
```
ColmTalbot, line 286: This isn't necessarily a comment for this PR, but I'm a little concerned that this isn't failing. It suggests that the different backends aren't falling over when we ask to implicitly cast between them.

bilby/gw/detector/networks.py
```diff
@@ -398,31 +399,83 @@ def __init__(
...
+        unit_vector_y = current_ifo.geometry.unit_vector_along_arm("y")
+
+        vertex_geocentric = get_vertex_position_geocentric(np.array([current_ifo.latitude_radians,
+                                                                     current_ifo.longitude_radians,
+                                                                     current_ifo.elevation]))
+        next_vertex_geocentric = vertex_geocentric + current_ifo.length * 1000 * unit_vector_x
+        next_vertex_ellipsoid = get_vertex_position_ellipsoid(next_vertex_geocentric)
+        next_latitude_rad, next_longitude_rad, next_elevation = next_vertex_ellipsoid
+
+        rotation_vector = np.cross(unit_vector_x, unit_vector_y)
```
ColmTalbot, line 449: As with other cases, these should all be `xp`.

bilby/gw/detector/geometry.py
```diff
@@ -208,8 +208,8 @@ def vertex(self):
         array_like: A 3D array representation of the vertex
         """
         if not self._vertex_updated:
-            self._vertex = gwutils.get_vertex_position_geocentric(self._latitude, self._longitude,
-                                                                  self.elevation)
+            self._vertex = gwutils.get_vertex_position_geocentric(np.array([self._latitude, self._longitude,
```
ColmTalbot, line 211: I'm not sure the best way to find the appropriate array backend here, but I would guess we want this import at the top of the file

```python
from ...compat.utils import array_module
```

then here we would have

```python
            xp = array_module((self._latitude, self._longitude, self.elevation))
```

ColmTalbot (changes requested):
Thanks! Just a few comments about the array backend stuff.

PR #897, opened 2025-01-29
TST: change all likelihood and waveform tests to use more recent waveforms  (#791)

### LINKED ISSUE #791: Change all likelihood and waveform tests to use more recent waveforms
In GitLab by @git.ligo:aditya.vijaykumar on Feb 23, 2024, 15:57

Currently, tests mostly use IMRPhenomPv2/D and other older waveforms. We want to move towards having more recent waveforms in the test suite. 

(See discussion: https://git.ligo.org/lscsoft/bilby/-/merge_requests/1268#note_950758)

akjm99:


test/gw/likelihood/marginalization_test.py
```diff
@@ -231,8 +231,8 @@ def setUp(self):
 
         trial_roq_paths = [
             "/roq_basis",
-            os.path.join(os.path.expanduser("~"), "ROQ_data/IMRPhenomPv2/4s"),
```
GregoryAshton, line 234: Here, I don't think we have ROQ data for XPHM yet, so this will need to be reverted (and may be why the CI is failing).

test/gw/likelihood_test.py
```diff
@@ -283,8 +283,8 @@ def setUp(self):
         # Possible locations for the ROQ: in the docker image, local, or on CIT
         trial_roq_paths = [
             "/roq_basis",
-            os.path.join(os.path.expanduser("~"), "ROQ_data/IMRPhenomPv2/4s"),
```
GregoryAshton, line 286: Same as the case above

GregoryAshton: And applied to the comments as well

test/gw/likelihood_test.py
```diff
@@ -546,8 +546,8 @@ def test_rescaling(self):
...
```
GregoryAshton, line 549: And again.

test/gw/likelihood/marginalization_test.py
GregoryAshton: @mj-will I assume #897 was a "move towards" modern waveforms and it is okay to use the older waveforms for the ROQs to avoid having to figure out how to package them?

test/gw/likelihood_test.py
```diff
@@ -1132,8 +1132,8 @@ def create_likelihood_single_basis(self):
...
```
GregoryAshton, line 1135: And here

GregoryAshton (changes requested):
Overall looks good thanks @akjm99. Just need to revert the cases where it is using an ROQ.

test/gw/likelihood/marginalization_test.py
mj-will: Yes, I think that's fine. 

I'll note there is definitely ROQ data for XPHM on CIT but I don't think it's public (and is also huge).

mj-will:
@akjm99 it looks like a commit from `main` has made it into the branch and is showing in the diff. I don't think it's from a merge or rebase so I'm slightly unsure what's happened to the history. @GregoryAshton any ideas?

akjm99:
Hi @mj-will. I followed the contributing guidelines which said to fetch upstream. When I tried to push my changes to my fork, it then said there were conflicts and that I had to merge my new changes and merge/rebase my branch to main (?) (which I think I managed to do, at least I was able to push to my fork after that...)

mj-will:
> Hi @mj-will. I am not sure how that happened... I followed the contributing guidelines which said to pull upstream. When I tried to push my changes to my fork, it then said there were conflicts and that I had to merge my new changes and merge/rebase my branch to main (?) (which I think I managed to do, at least I was able to push to my fork after that...)

Hmm, I think the upstream changed between you initially making the branch and now (the commit is from another recent PR). I think you've followed the contributing guidelines correctly but we're missing a section on updating a PR with changes from main after it has been made. I've opened an issue to track this: https://github.com/bilby-dev/bilby/issues/901.

Given we'll squash, this probably doesn't matter in this case, and it's probably more work to fix than it's worth.

test/gw/source_test.py
```diff
@@ -325,7 +325,7 @@ def setUp(self):
             frequency_nodes_linear=fnodes_linear,
             frequency_nodes_quadratic=fnodes_quadratic,
             reference_frequency=50.0,
-            waveform_approximant="IMRPhenomPv2",
```
GregoryAshton, line 322: I think this line should be using `IMRPhenomPv2` since we are passing in the ROQ terms just above

mj-will:
Hi @akjm99 do you have capacity to have another look at this? Unfortunately, I can't see what the test failures are but hopefully it's an easy fix.

akjm99:
Hi @mj-will, apologies; I thought I had managed to fix this already! I think you should be able to see the tests again now. Some of the unit tests are failing, but I am unsure why. I have a lot of work at the moment, but I will have another go at trying to debug this after I am back from the LVK.

PR #899, opened 2025-02-03
ENH: Added the CBCResult class to all gw examples

MattiaEmma2022:
Added a line in each result of CBC sampling occurrences to specify the result_class to be CBCResult

### DIFF
--- a/examples/gw_examples/data_examples/GW150914.py
+++ b/examples/gw_examples/data_examples/GW150914.py
@@ -109,5 +109,6 @@ result = bilby.run_sampler(
     check_point_plot=True,
     npool=1,
     conversion_function=bilby.gw.conversion.generate_all_bbh_parameters,
+    result_class=bilby.gw.result.CBCResult,
 )
 result.plot_corner()
--- a/examples/gw_examples/data_examples/GW190425.py
+++ b/examples/gw_examples/data_examples/GW190425.py
@@ -138,5 +138,6 @@ result = bilby.run_sampler(
     check_point_plot=True,
     npool=1,
     conversion_function=bilby.gw.conversion.generate_all_bns_parameters,
+    result_class=bilby.gw.result.CBCResult,
 )
 result.plot_corner()
--- a/examples/gw_examples/injection_examples/australian_detector.py
+++ b/examples/gw_examples/injection_examples/australian_detector.py
@@ -141,6 +141,7 @@ result = bilby.run_sampler(
     injection_parameters=injection_parameters,
     outdir=outdir,
     label=label,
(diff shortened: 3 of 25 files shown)


GregoryAshton:
@MattiaEmma2022 it looks like the `pre-commit` is failing. To fix this you can

1. Install `pre-commit` locally with `pip install pre-commit`
2. Run`pre-commit run --all-files` on this branch, this will make the necessary changes
3. Add the files and commit them

If you want to avoid this process in future, run `pre-commit install`.

GregoryAshton:
> Thanks for this Mattia! Looks good to me, however, I think the example in `supernova_example` needs updating as well (it uses the GW likelihood).

Does it make sense to use a CBCResult for a supernova analysis? I suspect that will cause issues as the CBCResult assumes the existence of certain parameters.

mj-will:
> Does it make sense to use a CBCResult for a supernova analysis? I suspect that will cause issues as the CBCResult assumes the existence of certain parameters.

It might be worth looking at this in more detail. I was thinking of all the interferometer, cosmology and waveform arguments which I think will still work but there may be some others that work. That said, it does use the `GraviationalWaveTransient` likelihood that has all the marginalizations. 

From the name though, I suppose it doesn't make sense to use it here. Perhaps this is an indication we should have a base `GWResult` from which `CBCResult` inherits?

GregoryAshton:
> > Does it make sense to use a CBCResult for a supernova analysis? I suspect that will cause issues as the CBCResult assumes the existence of certain parameters.
> 
> It might be worth looking at this in more detail. I was thinking of all the interferometer, cosmology and waveform arguments which I think will still work but there may be some others that work. That said, it does use the `GraviationalWaveTransient` likelihood that has all the marginalizations.
> 
> From the name though, I suppose it doesn't make sense to use it here. Perhaps this is an indication we should have a base `GWResult` from which `CBCResult` inherits?

Maybe, but I would suggest this is getting into the territory of "A foolish consistency is the hobgoblin of little minds" - i.e. I think in practise a GWResult with CBCResult subclass would be the correct way to implement it, but I don't think it is worthwhile (just my 2c).

ColmTalbot (commented):
This looks good to me, I'm happy to leave the supernova example out for now.

Did you run these examples, I'm always aware that we don't run at least some of these with any regularity.

MattiaEmma2022:
@ColmTalbot I have run all the core_examples and they run and complete with no problems

MattiaEmma2022:
I run a few of the gw examples and they run without errors and produce sensible results.

mj-will (approved):

ColmTalbot (approved):

PR #900, opened 2025-02-03
BUG: use likelihood for injection conversion function

ColmTalbot:
Resolves https://github.com/bilby-dev/bilby/issues/606

There has been a very long-standing issue where the `generate_all_cbc_parameter` functions didn't use the correct waveform arguments for injections. I think will resolve that in a generic way as long as the conversion function extracts that information from the likelihood. This now mirrors how the call is made for the posterior samples.

### DIFF
--- a/bilby/core/sampler/__init__.py
+++ b/bilby/core/sampler/__init__.py
@@ -335,7 +335,10 @@ def run_sampler(
             result.save_to_file(extension=save, gzip=gzip, outdir=outdir)
 
     if None not in [result.injection_parameters, conversion_function]:
-        result.injection_parameters = conversion_function(result.injection_parameters)
+        result.injection_parameters = conversion_function(
+            result.injection_parameters,
+            likelihood=likelihood,
+        )
 
     # Check if the posterior has already been created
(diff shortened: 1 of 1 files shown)


mj-will:
Would this mean we also need to update `identity_map_conversion` [here](https://github.com/bilby-dev/bilby/blob/395b90d62458841e71925532a7a39e0264ccf92b/bilby/gw/conversion.py#L2560)?

It might not cause issues immediately, but I think it would break `bilby_pipe` once/if [this MR](https://git.ligo.org/lscsoft/bilby_pipe/-/merge_requests/610) is merged.

mj-will:
> I don't think so, this is the "generation" function, so [this](https://github.com/bilby-dev/bilby/blob/395b90d62458841e71925532a7a39e0264ccf92b/bilby/gw/conversion.py#L2567) is the relevant function.
> 
> I just updated the docstring to specify this.

Ah, I'd missed the there are two functions. Thanks for pointing that out.

mj-will (approved):

adivijaykumar (approved):

Issue #908, opened 2025-02-07
Naming inconsistency for numerical relativity file


lorenzopompili00:
The waveform_kwarg to specify a numerical relativity file has been changed from `numerical_relativity_file` to `numerical_relativity_data` ([here](https://github.com/bilby-dev/bilby/blob/main/bilby/gw/source.py#L520)) in this commit https://github.com/bilby-dev/bilby/commit/768e1e049fde7654d457dc36364a7e632c6ed28e

I'm not sure if this was intentional. However, this is inconsistent with `bilby_pipe` which sets a key named `numerical_relativity_file` [here](https://git.ligo.org/lscsoft/bilby_pipe/-/blob/master/bilby_pipe/input.py?ref_type=heads#L433), and does not allow to setup e.g. NR injections with `bilby_pipe`

This is only true for the standard LAL interface, the gwsignal interface uses `numerical_relativity_file` [here](https://github.com/bilby-dev/bilby/blob/main/bilby/gw/source.py#L201)

PR #909, opened 2025-02-07
Restore numerical_relativity_file kwarg

lorenzopompili00:
Fixes https://github.com/bilby-dev/bilby/issues/908

Restores the waveform_kwarg to specify a numerical relativity file to be `numerical_relativity_file` rather than `numerical_relativity_data` in order to be consistent with [`bilby_pipe`](https://git.ligo.org/lscsoft/bilby_pipe/-/blob/master/bilby_pipe/input.py?ref_type=heads#L433).

Let me know if you prefer to use `numerical_relativity_data` instead and fix this in `bilby_pipe`.


ColmTalbot (approved):

mj-will:
I'm going to add this to 2.5.0 since I think it's a simple enough change that we can get it in.

mj-will (approved):

lorenzopompili00:
Thanks! Just rebased @mj-will

PR #911, opened 2025-02-10
BUG: fix incorrectly formatted raise statement

mj-will:
The raise statement in `get_all_injection_credible_levels` is incorrectly formatted. This PR fixes the formatting.

This wasn't being caught by the existing test, since the error was also a `TypeError`:

```
TypeError: exceptions must derive from BaseException
```

so I've also updated the test to check error message itself.


adivijaykumar (approved):

ColmTalbot (approved):
Good catch on the test failure.

PR #912, opened 2025-02-11
ENH: add `num_interp` to `AlignedSpin` prior

mj-will:
https://github.com/bilby-dev/bilby/pull/849 added a new test to the `prior_test.py` but this test is currently taking ~1 hour total in the CI. To address this, I've added `num_interp` to the `AlignedSpin` prior. 

This allows us to speed up the test but allows the user to configure the interpolation. The defaults are the same (different for each case).

On my laptop, this reduced the time from ~20 minutes to ~ 2 minutes, but we'll see how the CI behaves.

### DIFF
--- a/bilby/gw/prior.py
+++ b/bilby/gw/prior.py
@@ -487,9 +487,18 @@ class AlignedSpin(Interped):
     analytic form of the PDF.
     """
 
-    def __init__(self, a_prior=Uniform(0, 1), z_prior=Uniform(-1, 1),
-                 name=None, latex_label=None, unit=None, boundary=None,
-                 minimum=np.nan, maximum=np.nan):
+    def __init__(
+        self,
+        a_prior=Uniform(0, 1),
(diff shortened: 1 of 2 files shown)


adivijaykumar:
I haven't thought too much about this, but will the default of `10_000` and `100_000` also make sampling slower? We want to avoid that too, right?

mj-will:
> I haven't thought too much about this, but will the default of `10_000` and `100_000` also make sampling slower? We want to avoid that too, right?

So neither should affect sampling, it should be a one-time cost. Even with 100,000 the simple method seems to be quick enough that it doesn't slow down the tests.

Though perhaps it would be good to document these choices, @ColmTalbot was there a particular motivation behind these choices?

ColmTalbot (approved):

adivijaykumar (approved):

mj-will:
Thanks both, I'll probably add this in to 2.5.0 just to make preparing the release slightly quicker.

PR #915, opened 2025-02-17
MAINT: rework `GlobalMetaData` to inherit from `dict`

mj-will:
Rework `GlobalMetaData` to inherit from `dict` rather than `UserDict`.

This means any checks for `isinstance(o, dict)` will return `True` rather than `False`. This will hopefully address the issue seen in some downstream packages.

I've also added an environment variable called `BILBY_INCLUDE_GLOBAL_META_DATA` (open to suggestions for a better name) that toggles whether the meta data is included or not in the result object. By default, this is False, meaning it should be backwards compatible.

I also had to tweak the `cosmology` property in the CBC result now that it isn't always present.

### DIFF
--- a/bilby/core/result.py
+++ b/bilby/core/result.py
@@ -503,7 +503,7 @@ class Result(object):
         self.prior_values = None
         self._kde = None
 
-        if string_to_boolean(os.getenv("BILBY_INCLUDE_GLOBAL_METADATA", "False")):
+        if not string_to_boolean(os.getenv("BILBY_INCLUDE_GLOBAL_META_DATA", "False")):
             gmd = self.meta_data.pop("global_meta_data", None)
             if gmd is not None:
                 logger.warning(
@@ -511,6 +511,8 @@ class Result(object):
                     "Use the `BILBY_INCLUDE_GLOBAL_METADATA` environment variable to include it. "
                     "This behaviour will be removed in a future release."
                 )
+        else:
+            logger.debug("Including global meta data in the result object.")
 
     _load_doctstring = """ Read in a saved .{format} data file
 
--- a/bilby/core/result.py
+++ b/bilby/core/result.py
@@ -25,6 +25,7 @@ from .utils import (
     recursively_decode_bilby_json,
     safe_file_dump,
     random,
+    string_to_boolean,
 )
 from .prior import Prior, PriorDict, DeltaFunction, ConditionalDeltaFunction
 
@@ -501,6 +502,15 @@ class Result(object):
 
         self.prior_values = None
         self._kde = None
+
+        if string_to_boolean(os.getenv("BILBY_INCLUDE_GLOBAL_METADATA", "False")):
+            gmd = self.meta_data.pop("global_meta_data", None)
+            if gmd is not None:
+                logger.warning(
+                    "Global meta data was removed from the result object for compatibility. "
+                    "Use the `BILBY_INCLUDE_GLOBAL_METADATA` environment variable to include it. "
+                    "This behaviour will be removed in a future release."
+                )
 
     _load_doctstring = """ Read in a saved .{format} data file
 
--- a/bilby/core/result.py
+++ b/bilby/core/result.py
@@ -506,7 +506,7 @@ class Result(object):
         if not string_to_boolean(os.getenv("BILBY_INCLUDE_GLOBAL_META_DATA", "False")):
             gmd = self.meta_data.pop("global_meta_data", None)
             if gmd is not None:
-                logger.warning(
+                logger.info(
                     "Global meta data was removed from the result object for compatibility. "
                     "Use the `BILBY_INCLUDE_GLOBAL_METADATA` environment variable to include it. "
                     "This behaviour will be removed in a future release."
--- a/bilby/core/result.py
+++ b/bilby/core/result.py
@@ -509,7 +509,8 @@ class Result(object):
                 logger.info(
                     "Global meta data was removed from the result object for compatibility. "
                     "Use the `BILBY_INCLUDE_GLOBAL_METADATA` environment variable to include it. "
-                    "This behaviour will be removed in a future release."
+                    "This behaviour will be removed in a future release. "
+                    "For more details see: https://bilby-dev.github.io/bilby/faq.html#global-meta-data"
                 )
         else:
             logger.debug("Including global meta data in the result object.")
--- a/bilby/core/utils/meta_data.py
+++ b/bilby/core/utils/meta_data.py
@@ -1,16 +1,23 @@
-from collections import UserDict
-
 from . import random
 from .log import logger
 
 
-class GlobalMetaData(UserDict):
+class GlobalMetaData(dict):
     """A class to store global meta data.
 
     This class is a singleton, meaning that only one instance can exist at a time.
     """
 
     _instance = None
+
+    def __init__(self, mapping=None, /, **kwargs):
+        if mapping is None:
+            mapping = {}
+        else:
+            mapping = dict(mapping)
+        mapping.update(kwargs)
+        for key, item in mapping.items():
+            self.__setitem__(key, item)
 
     def __setitem__(self, key, item):
         if key in self:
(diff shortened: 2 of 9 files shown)


mj-will:
So this addresses the initial issue, but there are still issues saving the cosmology and RNG.

bilby/core/utils/meta_data.py
```diff
@@ -1,17 +1,24 @@
...
 
+    def __init__(self, mapping=None, /, **kwargs):
+        if mapping is None:
+            mapping = {}
+        else:
+            mapping = dict(mapping)
+        mapping.update(kwargs)
+        for key, item in mapping.items():
+            self.__setitem__(key, item)
+
```
asb5468, line 21: I don't quite see the use case for this. Can you give an example? I don't get how the simple example in the unit test will map to an actual bilby run.

asb5468 (commented):
Is the intention that `BILBY_INCLUDE_GLOBAL_METADATA` is an environment variable? (That's what I'm understanding from the added tests, at least.) Why not just make it a flag within bilby rather than relying on the environment variable being set?

bilby/core/result.py
```diff
@@ -502,6 +503,17 @@ def __init__(self, label='no_label', outdir='.', sampler=None,
         self.prior_values = None
         self._kde = None
 
+        if not string_to_boolean(os.getenv("BILBY_INCLUDE_GLOBAL_META_DATA", "False")):
+            gmd = self.meta_data.pop("global_meta_data", None)
+            if gmd is not None:
+                logger.warning(
```
ColmTalbot, line 509: I suggest making this info or lower when it isn't the default, at least for the first release.

Also, we should also probably add a small page to the docs that we can point to to avoid the barrage of "what does this mean" questions.

ColmTalbot (changes requested):

mj-will:
> Is the intention that `BILBY_INCLUDE_GLOBAL_META_DATA` is an environment variable? (That's what I'm understanding from the added tests, at least.) 

Yes, that's right. If we do go this route, then I should probably document this.

> Why not just make it a flag within bilby rather than relying on the environment variable being set?

I don't have a strong preference for an environment variable, I'm equally happy to rewrite this use a flag, but I'll explain my thought process. 

The main reason was to enable toggling this with the minimal number of changes to other functions in bilby and downstream codes. If it's flag in the `__init__` for the result class, then I think functions like [`read_in_result`](https://github.com/bilby-dev/bilby/blob/ef5871e88f55d65ecf2123117af1b8f4fb2e2ae7/bilby/core/result.py#L69) would need to be change to support the argument directly or `kwargs`. I also saw this as something that would be removed later and, to me, it's more hassle to add and remove a flag than an environment variable.

bilby/core/utils/meta_data.py
mj-will: So I based this directly on this post: https://realpython.com/inherit-python-dict/#inheriting-from-the-python-built-in-dict-class (that doesn't mean it's right).

My understanding is that `mapping` is needed to support things like this:

```python
f = dict({'one': 1, 'three': 3}, two=2)
```
which is supported by the default `dict` in python (see the docs [here](https://docs.python.org/3/library/stdtypes.html#mapping-types-dict)).

Looking at this now, I'm not sure if this is strictly needed.

bilby/core/result.py
mj-will: Good point, I'll make that change.

And yeah, you're right. I'll make a page to document this and also mention the environment variable (if we go with it).

mj-will: I've made changes to address both those points

docs/faq.txt
```diff
@@ -5,10 +5,26 @@ Frequently Asked Questions
...
+**A:** In `bilby` 2.5.0, the global meta data dictionary was added to the result object
+under :code:`result.meta_data["global_meta_data]`. This includes information such as the 
+global cosmology and random number generator. To ensure backwards compatibility,
+by default, this object is removed the result when it is instantiated.
+In a future release, this will be changed.
+In the meantime, you can include the global meta data by setting the
+global variable :code:`BILBY_INCLUDE_GLOBAL_META_DATA=1`. This can be either be done
+in the command line using :code:`export BILBY_INCLUDE_GLOBAL_META_DATA=1` or within
+python (e.g. using :code:`os`,
+code:`os.environ["BILBY_INCLUDE_GLOBAL_META_DATA"] = "1"`)
```
ColmTalbot, line 30:
```suggestion
:code:`os.environ["BILBY_INCLUDE_GLOBAL_META_DATA"] = "1"`)
```

docs/faq.txt
```diff
@@ -5,10 +5,26 @@ Frequently Asked Questions
...
 configuration of this. See the docstring of this :code:`bilby.core.utils.latex_plot_format`
 for the allowed configuration options.
 
+
+Global meta data
+----------------
+
+**Q:** I'm seeing a message about global meta data, what does this mean?
+
+**A:** In `bilby` 2.5.0, the global meta data dictionary was added to the result object
```
ColmTalbot, line 21:
```suggestion
**A:** In :code:`bilby` 2.5.0, the global meta data dictionary was added to the result object
```

bilby/core/result.py
```diff
@@ -502,6 +503,17 @@ def __init__(self, label='no_label', outdir='.', sampler=None,
         self.prior_values = None
         self._kde = None
 
+        if not string_to_boolean(os.getenv("BILBY_INCLUDE_GLOBAL_META_DATA", "False")):
+            gmd = self.meta_data.pop("global_meta_data", None)
+            if gmd is not None:
+                logger.info(
+                    "Global meta data was removed from the result object for compatibility. "
+                    "Use the `BILBY_INCLUDE_GLOBAL_METADATA` environment variable to include it. "
+                    "This behaviour will be removed in a future release."
```
ColmTalbot, line 512: I suggest directly adding a link to https://bilby-dev.github.io/bilby/faq.html#global-meta-data (link currently doesn't exist) to this message.

ColmTalbot (changes requested):

docs/faq.txt
```diff
@@ -5,10 +5,26 @@ Frequently Asked Questions
...
+
+Global meta data
+----------------
+
+**Q:** I'm seeing a message about global meta data, what does this mean?
+
+**A:** In :code:`bilby` 2.5.0, the global meta data dictionary was added to the result object
+under :code:`result.meta_data["global_meta_data]`. This includes information such as the 
+global cosmology and random number generator. To ensure backwards compatibility,
+by default, this object is removed the result when it is instantiated.
```
adivijaykumar, line 24: Some typo here? "this object is removed the result"

docs/faq.txt
```diff
@@ -5,10 +5,26 @@ Frequently Asked Questions
...
+In a future release, this will be changed.
+In the meantime, you can include the global meta data by setting the
+global variable :code:`BILBY_INCLUDE_GLOBAL_META_DATA=1`. This can be either be done
+in the command line using :code:`export BILBY_INCLUDE_GLOBAL_META_DATA=1` or within
+python (e.g. using :code:`os`,
+:code:`os.environ["BILBY_INCLUDE_GLOBAL_META_DATA"] = "1"`)
```
adivijaykumar, line 30: Maybe add explicitly what the behaviour is when BILBY_INCLUDE_GLOBAL_META_DATA is set to "0" or some other value? Don't have a strong feeling for this though, since I think the rest of the text makes what the behaviour should be obvious.

adivijaykumar (commented):
Looks mostly good to me! I am more in favour of making this an environment variable for the reason Colm mentioned, and also because the plan is for this to be temporary (at least as I understand it).

docs/faq.txt
mj-will: Fixed it, I think

mj-will:
@ColmTalbot @asb5468 @adivijaykumar it sounds like there is an overall preference for the environment variable approach. I'd like to get this in by the end of the week. So it would be great to get your thoughts on the changes I've made to address your feedback. Thanks!

ColmTalbot (approved):

adivijaykumar (approved):

mj-will:
@asb5468 I've merged this in so we can make a pre-release but I'm happy to make changes before the final release.

PR #914, opened 2025-02-13
CHANGELOG: update for v2.5.0

mj-will:
Update the changelog for v2.5.0.

PR #919, opened 2025-02-21
TST: reduce random test failures

mj-will:
As noted in https://github.com/bilby-dev/bilby/pull/914#issuecomment-2656692523, a specific test in the CI seems to be randomly failing, `test_reweight_different_likelihood_weights_correct`.

I ran this test locally 100 times using `parameterized` and it seemed to fail ~3% of the time. 

To try and avoid this impacting the CI, I've made two changes:

- Increase the number of samples to 2000, when I ran this locally it significantly reduced the number of failures (I had to run it x100 multiple times)
- Add a `pytest` mark that will rerun the test if it fails. This requires a new dependency which I've added to `optional_requirements.txt` and the containers environment.

### DIFF
--- a/containers/env-template.yml
+++ b/containers/env-template.yml
@@ -22,6 +22,7 @@ dependencies:
   - black
   - pytest-cov
   - pytest-requires
+  - pytest-rerunfailures
   - conda-forge::arviz
   - parameterized
   - scikit-image
--- a/optional_requirements.txt
+++ b/optional_requirements.txt
@@ -2,3 +2,4 @@ celerite
 george
 plotly
 pytest-requires
+pytest-rerunfailures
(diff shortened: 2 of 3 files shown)


mj-will:
> To be included in the images, it needs to be added to this [env file](https://github.com/bilby-dev/bilby/blob/main/containers/env-template.yml), that should be all that is needed for it to be automatically included when the images next rebuild.

Okay, I've done that. So hopefully it will update once this is merged.

GregoryAshton (approved):

adivijaykumar (approved):
Didn't know about pytest-rerunfailures! TIL.

adivijaykumar:
(also added the 2.5.0 milestone since we potentially want to get this change in?)

PR #920, opened 2025-02-24
ENH: Add InterferometerList time domain plotting

mick-wright:
Currently, the InterferometerList class has the `plot_data` function for plotting the frequency domain strain for each of the detectors directly using their respective `plot_data` functions. However, to plot the time domain equivalent, one must loop the detectors manually. This PR simply adds an equivalent function for the `plot_time_domain_data` function. Testing has been added that mirrors the `plot_data` testing.


mj-will (approved):
Thanks @mick-wright for adding this, LGTM.

PR #922, opened 2025-02-27
DOC:Fixed README link for opening bilby_pipe issues

mick-wright:
The link for opening bilby_pipe issues currently is broken as it points within the bilby-dec github, this MR restores it to point to the LVK gitlab.


mj-will (approved):
Ah, that dangers of find and replace. Thanks for spotting this Mick!

adivijaykumar (approved):

Issue #929, opened 2025-03-26
`generate_posterior_samples_from_marginalized_likelihood` converts sky frame parameters under the incorrect conditions


Rhiannon-Udall:
In `generate_posterior_samples_from_marginalized_likelihood` there is a conditional to call `generate_sky_frame_parameters` if the `time_reference!="geocent"` or `reference_frame!=sky`. However, [the conditional](https://github.com/bilby-dev/bilby/blob/65bcde32260ba9608d6fa2efee958020ee818f7e/bilby/gw/conversion.py#L1679) checks for "geocenter" not "geocent". While [the likelihood can be instantiated](https://github.com/bilby-dev/bilby/blob/main/bilby/gw/likelihood/base.py#L165) with `time_reference="geocenter"` (and indeed this is the default), the *property* `likelihood.time_reference` will always be either `geocent` or an ifo time. Accordingly, this conditional *always* runs, which in at least one case I was working on was erroneously setting sky parameters to be NaNs for some reason. Assuming this is the bug I think it is I can submit a quick fix @mj-will @ColmTalbot

ColmTalbot:
Sounds like a bug to me!

Rhiannon-Udall:
Alright, https://github.com/bilby-dev/bilby/pull/930 should fix this.

PR #930, opened 2025-03-27
BUG: Use "geocent" not "geocenter" where appropriate

Rhiannon-Udall:
In the conditional of `generate_posterior_samples_from_marginalized_likelihood`, checks for the value of `likelihood.time_reference` which will actually occur. Fixes https://github.com/bilby-dev/bilby/issues/929.

### DIFF
--- a/bilby/gw/conversion.py
+++ b/bilby/gw/conversion.py
@@ -1683,7 +1683,7 @@ def _generate_all_cbc_parameters(sample, defaults, base_conversion,
 
         if (
             not getattr(likelihood, "reference_frame", "sky") == "sky"
-            or not getattr(likelihood, "time_reference", "geocenter") == "geocenter"
+            or not getattr(likelihood, "time_reference", "geocent") == "geocent"
         ):
             try:
                 generate_sky_frame_parameters(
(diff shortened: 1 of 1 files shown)


mj-will:
Thanks for fixing this @Rhiannon-Udall. From the [doc-string](https://github.com/bilby-dev/bilby/blob/65bcde32260ba9608d6fa2efee958020ee818f7e/bilby/gw/likelihood/base.py#L96) I think the intention was to support both options rather than only `geocent`:

```
time_reference: str, optional
    Name of the reference for the sampled time parameter.

    - :code:`geocent`/:code:`geocenter`: sample in the time at the
      Earth's center, this is the default
    - e.g., :code:`H1`: sample in the time of arrival at H1
```

Could you update the PR to reflect this? It would also ensure the change is backwards compatible.

mj-will:
@Rhiannon-Udall do you think you'll be able to address the comment above in the next week or so? (We're keen to get this into the next release). If not, we may ask some to take this on if that's okay with you.

Rhiannon-Udall:
Sorry for the delay, I've made this change now @mj-will

mj-will (approved):
Thanks Rhiannon, LGTM.

(BTW, I like that we now support both spellings of centre :smile: )

ColmTalbot (approved):

PR #931, opened 2025-03-27
BUG: fix injection parameters being fixed during recovery

ColmTalbot:
#900 introduced a bug in the recovery of distance posteriors when using distance marginalization (among other things). This resolves the issue, but I'm not convinced that this is the best solution.

In general, I think this kind of situation is fairly unavoidable as we store the parameter values in the likelihood.

### DIFF
--- a/bilby/gw/conversion.py
+++ b/bilby/gw/conversion.py
@@ -1704,7 +1704,7 @@ def _generate_all_cbc_parameters(sample, defaults, base_conversion,
             logger.info(
                 "Generation of {} parameters failed with message {}".format(
                     key, e))
-    
+
     if likelihood is not None:
         likelihood.parameters.clear()
         likelihood.parameters.update(initial_likelihood_parameters)
--- a/bilby/gw/conversion.py
+++ b/bilby/gw/conversion.py
@@ -1730,7 +1730,7 @@ def generate_all_bbh_parameters(sample, likelihood=None, priors=None, npool=1):
 
     .. versionchanged:: 2.5.1
        To ensure that internal state of :code:`likelihood` is not changed by
-       this function. The initial value of :code:`likelihood.parameters` are
+       this function, the initial value of :code:`likelihood.parameters` are
        saved and reset at the end of the function.
     """
     waveform_defaults = {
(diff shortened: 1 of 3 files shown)


asb5468:
I think the better option is to write conversion functions that don't modify the likelihood in place. Why not just make a deepcopy of the likelihood after [this line](https://github.com/bilby-dev/bilby/blob/65bcde32260ba9608d6fa2efee958020ee818f7e/bilby/gw/conversion.py#L1635) instead?

isaaclegred:
Just to clarify, why does the value of the parameters in the likelihood matter for the call to `result.samples_to_posterior` below?

mj-will:
> I think the better option is to write conversion functions that don't modify the likelihood in place. Why not just make a deepcopy of the likelihood after [this line](https://github.com/bilby-dev/bilby/blob/65bcde32260ba9608d6fa2efee958020ee818f7e/bilby/gw/conversion.py#L1635) instead?

I'm curious about this as well, would this be possible or would it just be too slow since it would be constantly copying the data etc in the likelihood?

asb5468 (approved):

mj-will:
@ColmTalbot it would be good to try and get this is soon, if you have a chance, could you have a look at the pending questions?

bilby/gw/conversion.py
```diff
@@ -1715,6 +1727,11 @@ def generate_all_bbh_parameters(sample, likelihood=None, priors=None, npool=1):
         likelihood.interferometers.
     priors: dict, optional
         Dictionary of prior objects, used to fill in non-sampled parameters.
+
+    .. versionchanged:: 2.5.1
+       To ensure that internal state of :code:`likelihood` is not changed by
+       this function. The initial value of :code:`likelihood.parameters` are
```
adivijaykumar, line 1733: Seems like the `.` should be replaced by `,`?

bilby/gw/conversion.py
```diff
@@ -1747,6 +1764,10 @@ def generate_all_bns_parameters(sample, likelihood=None, priors=None, npool=1):
     npool: int, (default=1)
         If given, perform generation (where possible) using a multiprocessing pool
 
+    .. versionchanged:: 2.5.1
+       To ensure that internal state of :code:`likelihood` is not changed by
+       this function. The initial value of :code:`likelihood.parameters` are
```
adivijaykumar, line 1769: Same comment as above

mj-will (approved):
LGTM

bilby/gw/conversion.py
ColmTalbot:
```suggestion
       this function, the initial value of :code:`likelihood.parameters` are
```

bilby/gw/conversion.py
ColmTalbot:
```suggestion
       this function, the initial value of :code:`likelihood.parameters` are
```

adivijaykumar (approved):

mj-will (approved):

PR #933, opened 2025-04-03
Draft: Add a fisher-information-matrix sampler

GregoryAshton:
- Adds a sampler wrapper so that the `fisher` methods can be called using `run_sampler` for ease of use. 
- Adds a rejection sample method to reweight to the posterior
- Adds diagnostic plots
- Improve the maximum likelihood estimation method
- Replace the naive finite difference Hessian estimator with scipy.differentiate function


mj-will:
Closing since @GregoryAshton has confirmed this is superseded by https://github.com/GregoryAshton/bilby-laplace

PR #935, opened 2025-04-05
BUG Change the name of GEO from `GEO600` to `G1`

transientlunatic:
Apologies in advance if this is the wrong way to handle this.

When trying to analyse data contained in GEO frames, which have channel names prefixed with `G1` I ran into problems, as bilby_pipe is unable to understand (I think) a channel list where the channel isn't prefixed with the name of the detector as understood by bilby. So bilby was expecting to open GEO600:channel_name but this was stored in G1:channel_name.

I think it's possible this might be handled in bilby_pipe as well, though this approach does fix the problem.


mj-will:
@transientlunatic it would be nice to see this merged before the EoL for GEO. Do you have capacity to address Colm's comment above?

I also think there are some unrelated changes that shouldn't be in the PR.

transientlunatic:
@mj-will I'll take a look at this now. You're right that the ignore inconsistent stuff shouldn't be there!!!

transientlunatic:
@mj-will @ColmTalbot I've tidied this up a bit, though Claude may have been a bit excessive in engineering the backwards-compatibility, so if you want this simplified I would not be opposed to that.

ColmTalbot (approved):
I'm happy with this version. It isn't too verbose and I wouldn't be surprised if we need this again some time.

mj-will (approved):

PR #940, opened 2025-04-14
Cosmo prior loading

asb5468:
If bilby cannot find the conversion function used by the prior object stored inside a result file, it fails to `read_in_result` with the following error:
```
Traceback (most recent call last):                                                                                                                                                                                                  
  File "/home/sylvia.biscoveanu/O4_pop/spin_sorting/spin_sorting_generic.py", line 125, in <module>                                                                                                                                 
    main(**vars(args))                                                                                                                                                                                                              
  File "/home/sylvia.biscoveanu/O4_pop/spin_sorting/spin_sorting_generic.py", line 17, in main                                                                                                                                      
    data = bilby.core.result.read_in_result(result_file)                                                                                                                                                                            
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                            
  File "/home/sylvia.biscoveanu/.conda/envs/popsummary/lib/python3.11/site-packages/bilby/core/result.py", line 114, in read_in_result                                                                                              
    result = result_class.from_hdf5(filename=filename)                                                                                                                                                                              
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                              
  File "/home/sylvia.biscoveanu/.conda/envs/popsummary/lib/python3.11/site-packages/bilby/core/result.py", line 542, in from_hdf5                                                                                                   
    json.loads(data["priors"], object_hook=decode_bilby_json)                                                                                                                                                                       
  File "/home/sylvia.biscoveanu/.conda/envs/popsummary/lib/python3.11/json/__init__.py", line 359, in loads                                                                                                                         
    return cls(**kw).decode(s)     
               ^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                     
  File "/home/sylvia.biscoveanu/.conda/envs/popsummary/lib/python3.11/json/decoder.py", line 337, in decode                                                                                                                         
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())                                                                                                                                                                               
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                               
  File "/home/sylvia.biscoveanu/.conda/envs/popsummary/lib/python3.11/json/decoder.py", line 353, in raw_decode                                                                                                                     
    obj, end = self.scan_once(s, idx)                                                                                                                                                                                               
               ^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                               
  File "/home/sylvia.biscoveanu/.conda/envs/popsummary/lib/python3.11/site-packages/bilby/core/utils/io.py", line 183, in decode_bilby_json                                                                                         
    return getattr(import_module(dct["__module__"]), dct["__name__"], default)                                                                                                                                                      
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                 
  File "/home/sylvia.biscoveanu/.conda/envs/popsummary/lib/python3.11/importlib/__init__.py", line 126, in import_module                                                                                                            
    return _bootstrap._gcd_import(name[level:], package, level)                                                                                                                                                                     
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                     
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import                                                                                                                                                                   
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load                                                                                                                                                                
  File "<frozen importlib._bootstrap>", line 1140, in _find_and_load_unlocked                                                                                                                                                       
ModuleNotFoundError: No module named 'final_mass_models'
```
This MR attempts to catch this error and return the name of the function it's trying to find as a string instead. (Related to https://git.ligo.org/lscsoft/bilby/-/merge_requests/1360)


asb5468:
Will returning a string instead of a function break any other kind of object (besides prior) decoding? I don't know how else to test that if it's not caught by the test cases in the unit tests

asb5468:
The prior error message is somewhat unhelpful in this situation:
```
14:24 bilby WARNING : Cannot load module final_mass_models, returning function name as string
14:24 bilby WARNING : Unknown prior class for parameter None, defaulting to base Prior object
```
It's not that the `Prior` itself is unknown, just the conversion function. But the code to solve both errors is the same, so it didn't seem worth duplicating with another error message.

bilby/core/utils/io.py
```diff
@@ -291,7 +291,8 @@ def decode_bilby_json(dct):
     if dct.get("__prior__", False):
         try:
             cls = getattr(import_module(dct["__module__"]), dct["__name__"])
-        except AttributeError:
+            obj = cls(**dct["kwargs"])
+        except (AttributeError, ValueError):
             logger.warning(
                 "Unknown prior class for parameter {}, defaulting to base Prior object".format(
```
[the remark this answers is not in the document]
mj-will: @asb5468 just a gentle reminder about this. Do you have capacity to look at this before the call next week?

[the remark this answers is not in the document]
asb5468: I've updated this and confirmed that I now get the following error messages:
```
07:29 bilby WARNING : Cannot load module final_mass_models, returning function name as string
07:29 bilby WARNING : Unknown prior conversion function, defaulting to base Prior object
```

bilby/core/utils/io.py
```diff
@@ -291,19 +291,22 @@ def decode_bilby_json(dct):
...
+            obj = cls(**dct["kwargs"])
+        except (AttributeError, ValueError) as e:
+            if type(e).__name__ == 'AttributeError':
+                warning_message = "Unknown prior class for parameter {}, defaulting to base Prior object".format(
                     dct["kwargs"]["name"]
                 )
-            )
+            elif type(e).__name__ == 'ValueError':
+                warning_message = "Unknown prior conversion function, defaulting to base Prior object"
+            logger.warning(warning_message)
```
ColmTalbot, line 305: - Is this failure only triggered with the conversion function? I could imagine this also happening for the condition function for conditional priors, or am I missing something?
- The new warning message doesn't include the parameter that failed (`data["kwargs"]["name"]`), we should make sure to include that for debugging.

Googling around this, I found the `exc_info` option for the logger functions that prints the full traceback. Would it be overkill to include this?

asb5468: - `data["kwargs"]["name"]` returns `None` when this kind of exception is triggered, as there is no particular parameter associated with the conversion function failure. The conversion function is an attribute of the whole `PriorDict`.
- I do not know if this is trigged for the condition function for conditional priors. I only encountered it for the conversion function. If people have other example result files where this fails, I am happy to take those kinds of failures into account, but I don't think getting preemptively creative is a good use of time.
- Can you provide some example or resource for the `exc_info` option?

ColmTalbot: - My concern with this error being triggered by other mechanisms is that the current message is quite specific. If the error is not caused by a faulty conversion function, people will be confused by the message.
- The behaviour of `exc_info` is described in [this docstring](https://docs.python.org/3/library/logging.html#logging.Logger.debug).

ColmTalbot: Looks good now.

bilby/core/utils/io.py
```diff
@@ -291,19 +291,22 @@ def decode_bilby_json(dct):
...
-                "Unknown prior class for parameter {}, defaulting to base Prior object".format(
+            obj = cls(**dct["kwargs"])
+        except (AttributeError, ValueError) as e:
+            if type(e).__name__ == 'AttributeError':
+                warning_message = "Unknown prior class for parameter {}, defaulting to base Prior object".format(
                     dct["kwargs"]["name"]
                 )
-            )
+            elif type(e).__name__ == 'ValueError':
+                warning_message = f"Unable to load prior {cls} with arguments {dct['kwargs']}, defaulting to base Prior object"
```
ColmTalbot: This should fix the pre-commit issue.

```suggestion
                warning_message = (
                    f"Unable to load prior {cls} with arguments {dct['kwargs']}, "
                    "defaulting to base Prior object"
                )
```

Issue #937, opened 2025-04-11
Proposal: move away from parameters as state


ColmTalbot:
## tl;dr

Deprecate and remove the following pattern

```python
likelihood.parameters.update(parameters)
likelihood.log_likelihood()
```

in favour of

```python
likelihood.log_likelihood(parameters)
```

## Summary

Since the very early days we have had a `parameters` attribute of the `Likelihood` class that is used to evaluate the value of the likelihood.

I suggest we (eventually) remove this attribute and instead rely on users specifying the parameter values in the call to `Likelihood.log_likelihood`.

### Benefits

- Just-in-time (JIT) compilers (e.g., via `JAX`) typically require the variables for a function to be explicitly described. Currently formatting a likelihood for use with `JAX` requires a workaround such as [this](https://github.com/ColmTalbot/gwpopulation/blob/6a81fc7c79b59dd20254369012ce69e8589f6d7f/gwpopulation/experimental/jax.py#L8). The proposed change would make `bilby` likelihood calls natively JIT-compatible (the same argument probably applies for similar packages, e.g., `MLX`).
- Python 3.13 introduced a free-threaded build that enables thread-safe parallelism, rather than the current process-based parallelism. Thread-based parallelism can be more efficient in many cases, and allow easier use of shared memory. The current method of setting the `Likelihood.parameters` is not thread-safe as these values will typically vary between threads.
- I think it's better aesthetically, as above, calling the likelihood becomes a one-liner. It would also match better how we call the prior object.
- It is also safer for a few reasons. Currently, if you forget to specify some parameters, then the values already in the likelihood object will be used and can cause stealth errors. Also, the recent issue resolved in https://github.com/bilby-dev/bilby/pull/931 would not have happened, as this uses hidden values of parameters stored in the likelihood object.

### Stages

- Introduce the possibility of passing the `Likelihood.log_likelihood` (and similar) methods the parameters directly and fallback to the current behaviour (minor version increment)
- Add a deprecation warning for the current behaviour (minor version increment)
- Remove the current behaviour (major version increment)

### Implementation requirements

Initial
- [ ] add optional `parameters` argument to many methods in `Likelihood` and subclasses.
- [ ] add `parameters` to the `SamplingConvenienceDump`.
- [ ] update how samplers call the likelihood.
- [ ] update examples to demonstrate new behaviour.
- [ ] write documentation page with migration instructions.
- [ ] gw conversions will likely need updating.
- [ ] add tests with the new behaviour.

[...]

PR #941, opened 2025-04-15
ENH: Allow no parameters as state

ColmTalbot:
This PR will enable users to switch to directly calling

```python
likelihood.log_likelihood(parameters)
```

There are a small number of changes that I had to make that may impact users directly:
- the `Analytical1dLikelihood.model_parameters` is now a method.
- `PriorDict.fill_prior` has a pretty old default prior setting that I had honestly completely forgotten about. I have just unilaterally removed that, I can probably find a way to add it back, but I'd rather not have to.
- Some of the GW conversion functions, e.g., marginalized parameter reconstruction assume the new interface.

Deprecation steps:
- all subclasses of `Likelihood` that don't let `log_likelihood` and/or `log_likelihood_ratio` accept `parameters` will log a warning on import.
- there is a new environment variable that will allow us to make the switch easily for testing both with very loud warnings, and then straight errors whenever the `Likelihood.parameters` is accessed.

TODO
- [x] write documentation
- [x] add more testing of the explicit method
- [x] figure out what's happening with pymc

See #937

### DIFF
--- a/test/gw/source_test.py
+++ b/test/gw/source_test.py
@@ -615,11 +615,15 @@ class TestRelbinBBH(unittest.TestCase):
                 self.frequency_array, **raise_error_parameters
             )
 
-    def test_relbin_bbh_fails_without_fiducial_option(self):
-        with self.assertRaises(TypeError):
+    def test_relbin_bbh_runs_without_fiducial_option(self):
+        self.assertIsInstance(
             bilby.gw.source.lal_binary_black_hole_relative_binning(
-                self.frequency_array, **self.parameters
-            )
+                self.frequency_array,
+                **self.parameters,
+                **self.waveform_kwargs_binned,
+            ),
+            dict,
+        )
 
     def test_relbin_bbh_xpprecession_version(self):
         self.parameters.update(self.waveform_kwargs_fiducial)
@@ -692,10 +696,14 @@ class TestRelbinBNS(unittest.TestCase):
         )
 
     def test_relbin_bns_fails_without_fiducial_option(self):
-        with self.assertRaises(TypeError):
+        self.assertIsInstance(
             bilby.gw.source.lal_binary_neutron_star_relative_binning(
-                self.frequency_array, **self.parameters
-            )
+                self.frequency_array,
+                **self.parameters,
+                **self.waveform_kwargs_binned,
+            ),
+            dict,
+        )
 
     def test_fiducial_fails_without_tidal_parameters(self):
         self.parameters.pop("lambda_1")


GregoryAshton:
@ColmTalbot I very much support this! 

One possible addition that may help people is if `parameters` could also be an array ordered by the dict keys.

bilby/core/likelihood.py
```diff
@@ -1,13 +1,73 @@
...
+        super().__init_subclass__(**kwargs)
+        if (
+            "parameters" not in inspect.signature(cls.log_likelihood).parameters
+            or "parameters" not in inspect.signature(cls.log_likelihood_ratio).parameters
+        ):
+            logger.warning(
+                f"{cls} log_likelihood or log_likelihood_ratio method does not "
+                "accept 'parameters' as an argument. This is deprecated behaviour and will "
+                "be removed in a future release. See FIXME for more details."
+            )
```
mj-will: We should consider whether this should be a logger call or actually raise a warning e.g. a deprecation warning.

If we stick to the logger, we should make sure that the warning is raised even if the logger isn't configured with `setup_logger`

mj-will:
> @ColmTalbot I very much support this!
> 
> One possible addition that may help people is if `parameters` could also be an array ordered by the dict keys.

I think it would great if we can eventually support 'vectorized' likelihood functions. I think we should consider this when making  this change. I think we should consider different ways this could be implemented. A few come to mind:

- As Greg suggested, an array ordered by dict keys
- A dictionary with where each entry has length N
- Structured arrays (I'm not sure this is worth the headache since I don't think e.g. jax natively supports this)

Are there other options we should be considering?

GregoryAshton:
I agree with Colm that we shouldn't hold up this MR - we could convert the discussion above into an issue feature request.

One quick question - in the docs the version is listed as "XYZ" is that to be updated or will it be filled in automatically somehow?

bilby/core/likelihood.py
ColmTalbot: It looks like this is a case where we should use an actual warning.

bilby/core/prior/dict.py
```diff
@@ -341,25 +336,13 @@ def fill_priors(self, likelihood, default_priors_file=None):
             The filled prior dictionary
 
         """
+        if likelihood is not None:
+            warn("Filling priors from likelihood parameters is deprecated")
```
mj-will: I think these should be a specific type of warning. I think either `DeprecationWarning` or `FutureWarning`, the former isn't printed by default so I think I'd favour the latter. What do you think?

mj-will (changes requested):
Just one small comment and then I think this is good to go.

bilby/core/prior/dict.py
ColmTalbot: I went for a mix between whether they feel like downstream developer vs user issues. E.g., defining a subclass gets a `DeprecationWarning` as that is a downstream developer thing, but calling the likelihood without passing parameters gets a `FutureWarning` as users are likely to do that directly.

mj-will (approved):
LGTM

examples/gw_examples/injection_examples/relative_binning.py
```diff
@@ -81,6 +81,7 @@
 ifos.inject_signal(
     waveform_generator=waveform_generator, parameters=injection_parameters
 )
+injection_parameters["fiducial"] = 0
```
ColmTalbot: This change is actually needed on main after the recent changes to have the conversion function run on injected parameters, it is not directly related to this PR.

bilby/gw/likelihood/relative.py
```diff
@@ -234,10 +235,11 @@ def setup_bins(self):
 
     def set_fiducial_waveforms(self, parameters):
         parameters = parameters.copy()
-        parameters["fiducial"] = 1
+        self._set_fiducial()
         parameters.update(self.get_sky_frame_parameters(parameters=parameters))
         self.fiducial_polarizations = self.waveform_generator.frequency_domain_strain(
             parameters)
+        self._unset_fiducial()
```
mj-will, line 242: Why is this `unset` needed here? I don't think it was there before.

examples/gw_examples/injection_examples/relative_binning.py
```diff
@@ -49,14 +49,14 @@
...
     dec=-1.2108,
-    fiducial=1,
 )
 
 # Fixed arguments passed into the source model
 waveform_arguments = dict(
     waveform_approximant="IMRPhenomXP",
     reference_frequency=50.0,
     minimum_frequency=minimum_frequency,
+    fiducial=1,
```
mj-will, line 59: Is this still needed after adding the `set`/`unset` methods?

ColmTalbot: Yeah, this is needed for the injection

bilby/gw/likelihood/relative.py
ColmTalbot: These methods are new, so won't have been there before. This is needed for two things, unsetting the parameters cached in the waveform generator, and making sure the likelihood will use the heteronyned mode.

GregoryAshton (approved):

PR #945, opened 2025-04-17
MAINT: remove unsupported roq json weight file format

ColmTalbot:
This was slated for v2.2, but I guess we missed that...

### DIFF
--- a/bilby/gw/likelihood/roq.py
+++ b/bilby/gw/likelihood/roq.py
@@ -990,7 +990,7 @@ class ROQGravitationalWaveTransient(GravitationalWaveTransient):
     def save_weights(self, filename, format='hdf5'):
         """
         Save ROQ weights into a single file.
-        Support for json format was removed in :code:`v2.6`, only hdf5 and npz are supported.
+        Support for json format was removed in :code:`v2.7`, only hdf5 and npz are supported.
 
         Parameters
         ==========
--- a/bilby/gw/likelihood/roq.py
+++ b/bilby/gw/likelihood/roq.py
@@ -1,10 +1,7 @@
 
-import json
-
 import numpy as np
 
 from .base import GravitationalWaveTransient
-from ...core.utils import BilbyJsonEncoder, decode_bilby_json
 from ...core.utils import (
     logger, create_frequency_series, speed_of_light, radius_of_earth
 )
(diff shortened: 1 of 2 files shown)


bilby/gw/likelihood/roq.py
```diff
@@ -992,84 +989,54 @@ def _set_weights_quadratic_multiband(self, quadratic_matrix, basis_idxs):
 
     def save_weights(self, filename, format='hdf5'):
         """
-        Save ROQ weights into a single file. format should be npz, or hdf5.
-        For weights from multiple bases, hdf5 is only the possible option.
-        Support for json format is deprecated as of :code:`v2.1` and will be
-        removed in :code:`v2.2`, another method should be used by default.
+        Save ROQ weights into a single file.
+        Support for json format was removed in :code:`v2.5`, only hdf5 is supported.
```
ColmTalbot, line 993: Update to 2.6 for removal

bilby/gw/likelihood/roq.py
```diff
@@ -1078,8 +1057,6 @@ def load_weights(self, filename, format=None):
         """
         if format is None:
             format = filename.split(".")[-1]
-        if format not in ["json", "npz", "hdf5"]:
-            raise IOError(f"Format {format} not recognized.")
```
asb5468, line 1082: I can't find the suggest option, but what's the reason for removing this altogether instead of just reming the json option from the first line?

test/gw/likelihood_test.py
```diff
@@ -1046,18 +1046,22 @@ def test_from_npy(self, from_array):
 @pytest.mark.requires_roqs
 class TestInOutROQWeights(unittest.TestCase):
 
-    @parameterized.expand(['npz', 'hdf5'])
-    def test_out_single_basis(self, format):
```
asb5468, line 1050: Why are these tests only for hdf5 if we're keeping the npz functionality?

bilby/gw/likelihood/roq.py
ColmTalbot: I reverted this change.

test/gw/likelihood_test.py
ColmTalbot: I reverted this change.

bilby/gw/likelihood/roq.py
```diff
@@ -992,32 +989,22 @@ def _set_weights_quadratic_multiband(self, quadratic_matrix, basis_idxs):
 
     def save_weights(self, filename, format='hdf5'):
         """
-        Save ROQ weights into a single file. format should be npz, or hdf5.
-        For weights from multiple bases, hdf5 is only the possible option.
-        Support for json format is deprecated as of :code:`v2.1` and will be
-        removed in :code:`v2.2`, another method should be used by default.
+        Save ROQ weights into a single file.
+        Support for json format was removed in :code:`v2.6`, only hdf5 and npz are supported.
```
mj-will: This should be 2.7
```suggestion
        Support for json format was removed in :code:`v2.7`, only hdf5 and npz are supported.
```

mj-will (changes requested):
LGTM, just a version that needs bumping.

bilby/gw/likelihood/roq.py
```diff
@@ -1058,18 +1041,14 @@ def save_weights(self, filename, format='hdf5'):
 
     def load_weights(self, filename, format=None):
         """
-        Load ROQ weights. format should be json, npz, or hdf5.
-        json or npz file is assumed to contain weights from a single basis.
-        Support for json format is deprecated as of :code:`v2.1` and will be
-        removed in :code:`v2.2`, another method should be used by default.
+        Load ROQ weights. Support for json format was removed in :code:`v2.6`.
```
mattpitkin: Should this line now be updated to v2.7?

ColmTalbot: I thought I fixed this the other day...

PR #949, opened 2025-04-23
MAINT: add dynesty pin

ColmTalbot:
There are some upcoming breaking changes in dynesty that we should protect against (https://github.com/joshspeagle/dynesty/pull/471).


adivijaykumar (approved):
The speed improvements in the dynesty PR show very nice speed improvements for the `unif` sampling. From what I understand, our sampling methods will not show these. Is that right?

mj-will (approved):

PR #950, opened 2025-04-25
ENH: add support for new dynesty api

ColmTalbot:
In preparation for https://github.com/joshspeagle/dynesty/issues/495 this would allow us to (temporarily) support both versions of dynesty. There's some duplication, but this means that when we want to stop supporting the older version of dynesty, we can just delete the old `dynesty_utils.py` and some of the other methods.

This will be a draft for a while until the implementation in dynesty is finalized.

### DIFF
--- a/bilby/core/sampler/dynesty.py
+++ b/bilby/core/sampler/dynesty.py
@@ -189,6 +189,14 @@ class Dynesty(NestedSampler):
         kwargs["seed"] = None
         return kwargs
 
+    @property
+    def new_dynesty_api(self):
+        try:
+            from dynesty.sampling import InternalSampler  # noqa
+            return True
+        except ImportError:
+            return False
+
     def __init__(
         self,
         likelihood,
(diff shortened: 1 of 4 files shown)


mj-will:
@ColmTalbot could post a diff of the old and new utils?

bilby/core/sampler/dynesty.py
```diff
@@ -189,6 +189,15 @@ def default_kwargs(self):
         kwargs["seed"] = None
         return kwargs
 
+    @property
+    def new_dynesty_api(self):
+        try:
+            import dynesty.internal_samplers  # noqa
+
```
mj-will, line 196: I think until this is reviewed we should print a warning that using newer versions of dynesty is not reviewed. What do you think?

I understand that this shouldn't change the sampling, but I think we should verify this with at least a subset of the review tests.

mj-will (commented):
Overall LGTM, just one comment on how we handle LVK review for this.

bilby/core/sampler/dynesty.py
ColmTalbot: What level of review/warning are we hoping for? We don't comment on any other samplers not having reviewed settings.

My preference would be to just run a bunch of tests once we have something close to a stable 2.7.0. If we get the major PRs in, we can make a RC and then run a couple of fiducial tests before making the full release, we can pretty easily do this for both dynesty versions. This will also be the most effective way to catch subtle bugs.

bilby/core/sampler/dynesty.py
```diff
@@ -793,8 +822,17 @@ def read_saved_state(self, continuing=False):
                 if continuing:
                     self._remove_live()
                 self.sampler.nqueue = -1
-                self.start_time = self.sampler.kwargs.pop("start_time")
-                self.sampling_time = self.sampler.kwargs.pop("sampling_time")
+                if hasattr(self.sampler, "_bilby_metadata"):
+                    extras = self.sampler._bilby_metadata
+                elif hasattr(self.sampler, "kwargs"):
```
GregoryAshton, line 827: I'm a bit confused here. Why can the sampler have either metadata or kwargs?

bilby/core/sampler/dynesty.py
GregoryAshton: One could imagine an INFO log-level message stating which sampler API is being used. However, I am leaning towards less print out messages so would not vote for this.

mj-will: As discussed on the call, we agreed this isn't needed.

mj-will (approved):

bilby/core/sampler/dynesty.py
ColmTalbot: With the new API the dynesty sampler classes don't have a `kwargs` attribute that we can piggyback on.

PR #952, opened 2025-05-01
BLD: migrate to pyproject.toml

mj-will:
This PR migrates the code from using a combination of  `setup.py`, `setup.cfg` and `pyproject.toml` to solely `pyproject.toml`.

`pyproject.toml` is now the preferred way to package Python code and, as I understand it, will be required for Python 3.13.

**Changes**

- Remove `setup.py` and `setup-cfg`
- Bump minimum setuptools version
- Update pre-commits to use `pyproject.toml`

**To do**:

- [ ] Decide who we list as author(s) (toml supports multiple authors)


fgittins:
Thanks for putting this together @mj-will! I think this is a positive change that the Python ecosystem is moving towards.

I noticed that there is a small level of redundancy across the various `requirements.txt` files, but it's not a major issue.

mj-will:
> Thanks for putting this together @mj-will! I think this is a positive change that the Python ecosystem is moving towards.
> 
> I noticed that there is a small level of redundancy across the various `requirements.txt` files, but it's not a major issue.

Yeah, I agree. I think we could look at revising some of the requirements (and possibly adding some version constraints) in another PR.

pyproject.toml
```diff
@@ -1,10 +1,122 @@
...
+Documentation = "https://bilby.readthedocs.io"
+Source = "https://github.com/bilby-dev/bilby"
+Tracker = "https://github.com/bilby-dev/bilby/issues"
+
+[project.scripts]
+bilby_plot = "cli_bilby.plot_multiple_posteriors:main"
+bilby_result = "cli_bilby.bilby_result:main"
+
+[project.entry-points."bilby.samplers"]
+"bilby.bilby_mcmc" = "bilby.bilby_mcmc.sampler:Bilby_MCMC"
```
raffienficiaud, line 38: nitpic: I would use a formatter for TOML: indenting the sections would be performed automatically and would add clarity.

.pre-commit-config.yaml
```diff
@@ -15,12 +20,16 @@ repos:
...
+        additional_dependencies:
+        - tomli
+-   repo: https://github.com/pycqa/isort
+    rev: 5.11.5
     hooks:
-    -   id: isort # sort imports alphabetically and separates import into sections
-        args: [-w=88, -m=3, -tc, -sp=setup.cfg ]
+      - id: isort
+        name: isort (python)
+        args: [-w=88, -m=3, -tc]
```
raffienficiaud, line 30: I am not completely sure, but I believe `isort` can read the config directly from pyproject.toml, also from pre-commit

pyproject.toml
mj-will: Good idea, do you have any suggestions for a specific formatter?

.pre-commit-config.yaml
mj-will: You're right, I've updated that. I've also switched to using the `profile` option which includes a set of defaults for `black` which match what we had: https://pycqa.github.io/isort/docs/configuration/profiles.html

pyproject.toml
mj-will: I've add this pre-commit hook: https://github.com/macisamuele/language-formatters-pre-commit-hooks?tab=readme-ov-file. It seems to work but doesn't indent this block.

mj-will:
> I'm happy to leave the authors section generic as it currently is and add a maintainers section that covers the people who regularly review PRs/attend calls.

I've added a set of maintainers based who's been reviewing recent PRs. Let me know what you think. I can also add emails if we want.

mj-will:
> We might want to make an alpha release or something as a test after this is merged while it is fresh in our minds. I'm not sure if there's a way to test the release infrastructure otherwise.

Yeah, I think that's a good idea.

raffienficiaud:
> > We might want to make an alpha release or something as a test after this is merged while it is fresh in our minds. I'm not sure if there's a way to test the release infrastructure otherwise.
> 
> Yeah, I think that's a good idea.

You can always use test.pypi.org to check the things you are developping and how pypi.org would handle it. This works exactly like the official pypi, however the project name should be different (meaning also changing the toml file locally and regenerate the artifacts for the sake of testing).

pyproject.toml
raffienficiaud: Interesting! I was using some formatter in VScode "Even better TOML formatter".

pyproject.toml
```diff
@@ -1,10 +1,149 @@
...
+maintainers = [
+  {name = "Greg Ashton"},
+  {name = "Sylvia Biscoveanu"},
+  {name = "Colm Talbot"},
+  {name = "Aditya Vijaykumar"},
+  {name = "Michael J. Williams"}
+]
+name = "bilby"
+readme = "README.rst"
+requires-python = ">=3.10"
```
fgittins, line 32: This made me wonder: do we support Python 3.13?

pyproject.toml
```diff
@@ -1,10 +1,149 @@
...
+
+[tool.setuptools.dynamic.optional-dependencies]
+all = {file = [
+  "gw_requirements.txt",
+  "mcmc_requirements.txt",
+  "sampler_requirements.txt",
+  "optional_requirements.txt"
+]}
+gw = {file = ["gw_requirements.txt"]}
+mcmc = {file = ["mcmc_requirements.txt"]}
```
fgittins, line 141: This is obviously fine and works. However, it is more standard to write the dependencies explicitly in the `pyproject.toml`.

pyproject.toml
```diff
@@ -1,10 +1,149 @@
...
+  "past",
+  "pycbc",
+  "pymc3",
+  "pytest",
+  "scipy",
+  "setuptools",
+  "skimage",
+  "torch"
+]
+profile = "black"
```
fgittins, line 101: This might be better addressed in another PR, but I have recently been using [`ruff`](https://github.com/astral-sh/ruff) for my Python linting, which is way faster than other tools on the market. In addition to its speed, it is highly customisable.

It can be (more or less) used as a drop-in replacement for `black`, `flake8` and `isort`.

pyproject.toml
fgittins, line 1: Given that we are moving over to `pyproject.toml`, we may want to make use of [dependency groups](https://peps.python.org/pep-0735/).

For example, I tend to have a `dev` group, which includes the tools that I use during development that I don't want shipped with the package, such as `ruff`, `pytest` and `mypy`. This is sufficient for me to properly lint and format my code and test it locally.

fgittins (commented):
LGTM! I've left a few suggestions and comments to mull over.

pyproject.toml
mj-will: Last time I checked, I think we were limited by LALSuite: https://git.ligo.org/lscsoft/lalsuite/-/issues/778

pyproject.toml
mj-will: Yeah, I'd be in favour of that is well. I suggest we make an issue for it.

pyproject.toml
mj-will: I hadn't come across dependency groups before. I think we've been using `optional-dependencies` for some of this, is there a strong reason to use one over the other?

pyproject.toml
mj-will: I don't have a strong preference about this but I suggest we consider that in a follow-up. I know some people like using those files.

pyproject.toml
fgittins: The central idea is that the dependency groups are *not* included in the built distribution on, say, PyPI. So if we recommend linting with `flake8`, then this is the natural place to enforce that in the repo. Not in the `optional-dependencies` which can be installed by an end user with `pip install "bilby[gw]"`.

What is appealing to me as a developer is that I can clone the repo, generate a virtual environment and then the `pyproject.toml` automatically provides me the necessary tools to test and start writing code, according to the repo's standards and practices.

Note: this approach is also adopted by `scipy` (see their [`pyproject.toml`](https://github.com/scipy/scipy/blob/main/pyproject.toml#L72)).

pyproject.toml
fgittins: That's a fair point. Fortunately, the `optional-dependencies` can be installed in a similar fashion:

```
pip install -r gw_requirements.txt
```

becomes

```
pip install .[gw]
```

raffienficiaud: It might be just a matter of taste, but I like 

* having as few files as possible on the base of the repository
* one single place where I manage the dependencies (meaning I got rid of all `requirements` type of files)
* knowing why I am installing the dependencies for: `develop` or `tests`, `doc`, `pycbc_plugins` etc

ColmTalbot: I could be convinced to remove these files, I agree that it is aesthetically neater, but I think there are caveats.

> That's a fair point. Fortunately, the `optional-dependencies` can be installed in a similar fashion:
> 
> ```
> pip install -r gw_requirements.txt
> ```
> 
> becomes
> 
> ```
> pip install .[gw]
> ```

These aren't actually the same as the latter will also install the package, there are some cases where just the dependencies are desired. Also, if the goal is to install with conda then you can `conda install --file gw_requirements.txt`.

pyproject.toml
ColmTalbot: Bilby can be used with 3.13, but some of the optional extras won't work.

pyproject.toml
ColmTalbot: I'd be in favour of that, if we do globally refactor things we should try to set the blame to ignore that commit. I haven't used https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/ before, but it seems like it should be straightforward.

pyproject.toml
mj-will: I've opened an issue to track this: https://github.com/bilby-dev/bilby/issues/975

pyproject.toml
mj-will: I've open an issue to continue this discussion: https://github.com/bilby-dev/bilby/issues/976

Issue #944, opened 2025-04-17
Inefficient sampling of (efficient) conditional prior


fgittins:
When using conditional priors, I encounter the following warning message:

```
12:55 bilby WARNING : Prior sampling efficiency is very low, please verify its validity.
```

This is surprising, as the conditional prior I use is constructed to achieve a sampling efficiency of 100%. The upper bound of `mass_2` is always set to be less than or equal to `mass_1`, so all samples should be accepted.

This looks similar to the issue raised in #938.

### Minimal reproducible example

The following example defines a uniform prior for `mass_1`, and a conditional uniform prior for `mass_2` such that `mass_2 <= mass_1`:

```python
from bilby.core import prior

def condition_func(reference_params, mass_1):
    condition = min(reference_params["maximum"], mass_1)
    return dict(minimum=reference_params["minimum"], maximum=condition)

priors = prior.ConditionalPriorDict()
priors["mass_1"] = prior.Uniform(minimum=1.2, maximum=2)
priors["mass_2"] = prior.ConditionalUniform(
    minimum=1, maximum=1.4, condition_func=condition_func
)

samples = priors.sample(1_000_000)
```

I also observe that changing the bounds (even to make the condition redundant) does not affect the warning. For example:

```python
priors["mass_2"] = prior.ConditionalUniform(
    minimum=1, maximum=1.1, condition_func=condition_func
)
```

still yields the same warning.

### System information

- Bilby v2.5.0
- Python v3.10.16

Any insight into why this warning appears would be appreciated!

fgittins:
That was my initial suspicion, but wanted to check.

Issue #938, opened 2025-04-12
False warning of prior sampling efficiency issue


HuiTong5:
I got this [warning](https://github.com/bilby-dev/bilby/blob/65bcde32260ba9608d6fa2efee958020ee818f7e/bilby/core/prior/dict.py#L452) mistakenly. I didn't have any constraint for the parameters in my prior dict so the efficiency should be 100%.
It turned out the problem comes from this [line](https://github.com/bilby-dev/bilby/blob/65bcde32260ba9608d6fa2efee958020ee818f7e/bilby/core/prior/dict.py#L474). When there is no constraint in a prior dict, the returned value of this [`evaluate_contraints`](https://github.com/bilby-dev/bilby/blob/65bcde32260ba9608d6fa2efee958020ee818f7e/bilby/core/prior/dict.py#L62) function is always 1 but not an array with the size of the samples. This results in an incorrect `np.sum(keep)=1`.

mj-will:
Thanks for tracking this down @HuiTong5! Seems like this should be an easy fix.

mj-will:
This was fixed in https://github.com/bilby-dev/bilby/pull/953

PR #953, opened 2025-05-01
BUG: Fix sampling efficiency warning

fgittins:
This pull request fixes the bug reported in #938 and #944.

The issue is that, even for 100% efficient priors, like

```python
priors = bilby.core.prior.PriorDict()
priors["mass_1"] = bilby.core.prior.Uniform(minimum=1.2, maximum=2)
```

the following warning shows up when drawing a sufficiently large number of samples [here `priors.sample(1_000_000)`]:

```
21:25 bilby WARNING : Prior sampling efficiency is very low, please verify its validity.
```

The warning materialises because the `sample_subset_constrained` method of `PriorDict` assumes that the priors include at least one `Constraint` prior.

I have also added a very simple unit test to check for this warning.

### DIFF
--- a/bilby/core/prior/dict.py
+++ b/bilby/core/prior/dict.py
@@ -441,6 +441,23 @@ class PriorDict(dict):
         return [k for k, p in self.items() if isinstance(p, Constraint)]
 
     def sample_subset_constrained(self, keys=iter([]), size=None):
+        """
+        Sample a subset of priors while ensuring constraints are satisfied.
+
+        Parameters
+        ==========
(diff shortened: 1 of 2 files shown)


mj-will (approved):
Thanks @fgittins, this looks good to me!

adivijaykumar (approved):
Thanks, looks good. I didn't know about `patch`, it seems to be very useful!

mj-will:
@fgittins could you rebase this so I can merge it?

PR #954, opened 2025-05-02
BLD: use release branches

mj-will:
Change how we make releases to use release branches.

This should make it easier to backport bug fixes to versions being used in production analyses.

Note: this requires https://github.com/bilby-dev/bilby/pull/952

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -340,11 +340,21 @@ When making a changelog keep the following in mind:
 - Include links to the relevant PRs
 - Remember to update the URLs at the bottom of changelog file
 
-### Tagging a release
+### Making the release branch
 
 #### Major and minor releases 
 
 If making a major or minor release you must first make the release branch from the current `main` branch
+
+##### Via the web interface
+
+From the [branches page of the repo](https://github.com/bilby-dev/bilby/branches), click `New branch`,
+select `main` as the source and name the branch `release/Major.Minor.x` for example `release/2.6.x`.
+
+##### Via the command line
+
+Ensure you have checked out the `main` branch and it is up-to-date with the upstream repo.
+Then run the following commands with the appropriate version number:
 
 ```bash
 git checkout -b release/2.6.x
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -324,6 +324,9 @@ The version tag should also start with `v` e.g. `v2.4.0`.
 `bilby` uses `setuptools_scm` to automatically set the version based on git tags.
 This means no manual changes are needed to the version number are required.
 
+As on version v2.6.0, `bilby` uses a [stable mainline](https://www.bitsnbites.eu/a-stable-mainline-branching-model-for-git/) branching model for releases.
```
ColmTalbot, line 327:
```suggestion
As of version v2.6.0, `bilby` uses a [stable mainline](https://www.bitsnbites.eu/a-stable-mainline-branching-model-for-git/) branching model for releases.
```

CONTRIBUTING.md
```diff
@@ -337,6 +340,40 @@ When making a changelog keep the following in mind:
 - Include links to the relevant PRs
 - Remember to update the URLs at the bottom of changelog file
 
+### Tagging a release
+
+#### Major and minor releases 
+
+If making a major or minor release you must first make the release branch from the current `main` branch
+
+```bash
```
ColmTalbot, line 349: I usually do these things through the GUI (making the release branch/tags/releases), is there a reason the command line version is being recommended instead?

The only thing to note is that the new tag should be made from the release branch and not from `main`.

ColmTalbot (commented):
I'm happy with this change, and the docs generally look good. I just had a question about the GUI vs CLI approaches.

CONTRIBUTING.md
mj-will: I tend to it via the command line so I can verify I've tagged the correct commit but I agree both methods work.

I've updated the instructions to mention both methods.

ColmTalbot: I'm just basic.

ColmTalbot (approved):

PR #960, opened 2025-05-27
BUG: fix the filename extension precedence for result files

### LINKED ISSUE #959: Result.save_to_file now always saves as json unless explicitly specified otherwise
After https://github.com/bilby-dev/bilby/pull/858 the file extension given in the filename is silently overwritten in `Result.save_to_file`. For example `bilby_object.save_to_file("test.hdf5")` produces a file `test.json`.

I think the current default is incorrect, we should change the default to `extension=None` to allow us to directly use the provided extension. The documentation also probably needs to be slightly updated.

Desired precedence order:

- `extension`
- `filename.ext` if `extension` is `None`

e.g.,

- `filename=foo.bar`, `extension=None`: `foo.bar`
- `filename=foo.bar`, `extension=json`: `foo.json`

ColmTalbot:
Closes #959 

This does slightly change the behaviour, but I think this gets us back to the spirit of the implementation.

### DIFF
--- a/bilby/core/result.py
+++ b/bilby/core/result.py
@@ -811,8 +811,8 @@ class Result(object):
                 extension = "json"
                 filename = base_filename
             filename = f"{base}.{extension}"
-            if extension is None or extension is True:
-                extension = 'json'
+        if extension is None or extension is True:
+            extension = 'json'
 
         outdir = _outdir if outdir is None else outdir
         outdir = self._safe_outdir_creation(outdir, self.save_to_file)


bilby/core/result.py
```diff
@@ -785,22 +784,27 @@ def save_to_file(self, filename=None, overwrite=False, outdir=None,
...
-            extension = "json"
-
         _outdir = None
         if filename is not None:
-            _outdir, filename = os.path.split(filename)
+            _outdir, base_filename = os.path.split(filename)
             _outdir = None if _outdir == "" else _outdir
-            filename = f"{os.path.splitext(filename)[0]}.{extension}"
+            base, ext = os.path.splitext(base_filename)
+            if extension is None or extension is True:
```
asb5468, line 800: Is this meant to catch backwards compatibility? Based on the edited documentation, `extension` should never be `True`.

ColmTalbot: Yeah, I guess I should add a test for this case. I'm not really sure why we chose `True` as a fallback value.

mattpitkin (approved):
Looks good to me.

PR #961, opened 2025-06-02
REL: backport #960 to v2.5.2

mj-will:
Back port #960 to v2.5.x to make v2.5.2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ The original MRs are only visible on the [LIGO GitLab repository](https://git.li
 
 ## [Unreleased]
 
+## [2.5.2]
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


ColmTalbot (approved):
I'm happy to merge this, can you do a test run of what version we get on main after this tag is made?

mj-will:
> > This won't happen once #954 is merged but in the meantime we'll have to decide how to handle this. I think we have two options:
> > 
> > * Accept the version will be wrong until we make the next tag
> > * Merge the release/v2.5.2 branch back in to main after the release has been made
> >   What do you think?
> 
> I think the latter here is better, I don't think we should delay the release too much trying to figure out the best way to do this.

Okay, sounds good to me.

mattpitkin (approved):

PR #962, opened 2025-06-10
MAINT: update file extension logic

mj-will:
Update the file extension logic in `Result.save_to_file` to match what is described in the doc-string.

The main change is to no longer replace or add an extension when `filename` is specified. For example `filename=test, extension='pkl'` will just save to `test` rather than `test.pkl`.

I also update `read_in_result` so that extension isn't overridden.

This consistent with versions <2.4.0 but differs to versions since.

I also added explicit support for `h5` and `pickle`.

**Changes to `read_in_result`**

- Previously, the final if/else statement couldn't actually handle `None`
- The extension was always overridden by the value inferred from the file. Now there is only a default (`.json`) if the filename is not provided. This means the behaviour when `filename=None` is identical. When `filename` is specified it is the behaviour is also the same but the error that is raised when `extension=None` and the filename is missing in extension is slightly different.

### DIFF
--- a/test/gw/detector/interferometer_test.py
+++ b/test/gw/detector/interferometer_test.py
@@ -4,6 +4,7 @@ from unittest import mock
 
 import lal
 import lalsimulation
+import pytest
 from shutil import rmtree
 
 import numpy as np
@@ -594,6 +595,7 @@ class TestInterferometerAntennaPatternAgainstLAL(unittest.TestCase):
                 self.assertAlmostEqual(std, 0.0, places=10)
 
 
+@pytest.mark.flaky(reruns=3, only_rerun=["AssertionError"])
 class TestInterferometerWhitenedStrain(unittest.TestCase):
     def setUp(self):
         self.duration = 64


bilby/core/result.py
```diff
@@ -794,24 +805,34 @@ def save_to_file(self, filename=None, overwrite=False, outdir=None,
...
+                if extension is None:
+                    extension = ext
+                elif ext != extension:
+                    message = (
+                        f"The specified extension '{ext}' "
+                        f"does not match the provided extension '{extension}'. "
+                    )
+                    logger.warning(message)
+
+        if extension is None:
```
ColmTalbot, line 854: Since we're adding logging, it would be good to add an info like "No file extension specified, defaulting to JSON."

ColmTalbot (commented):
This generally looks good. On testing, it would be good to add some tests of what happens if you attempt to read a file saved with one format with the wrong read_in_result extension.

mj-will:
> This generally looks good. On testing, it would be good to add some tests of what happens if you attempt to read a file saved with one format with the wrong read_in_result extension.

Good idea. Whilst adding these tests I realised it would probably be insightful to actually raise a helpful error, so I've slightly reworked the error handling in `read_in_result`.  Let me know what you think.

bilby/core/result.py
mj-will: Good point, I've added a couple more statements.

test/core/result_test.py
```diff
@@ -567,6 +602,8 @@ class NotAResult(bilby.core.result.Result):
         # result should be specified result_class
         assert isinstance(result, NotAResult)
 
+        # TODO: figure out why the caching message is not logged
```
ColmTalbot: There's a todo here, is the aim to resolve this later? Otherwise, this looks good to me.

mj-will: Thanks for the reminder. I've had another look at this and managed to fix it.

PR #967, opened 2025-06-26
BUG: remove incorrect window-based PSD scaling

### LINKED ISSUE #869: Optimal SNR Depends on Data Conditioning
This is likely not issue per se but may be a point about notation/definition.

When computing the optimal SNR using the routine [here](https://github.com/bilby-dev/bilby/blob/main/bilby/gw/detector/interferometer.py#L565) bilby uses the PSD array, which will include the `strain_data.window_factor` if the data were read in from a frame file and FFT'ed. This makes sense as a way to correct for the power loss in the noise specifically when computing the log likelihood, which calls the optimal SNR and matched filter SNR functions.

However, when using bilby to report the optimal SNR of signals as an output of an analysis, does it make sense for this to depend on how the *data* are conditioned given the optimal SNR is (h|h)? This may make sense if we are scaling the PSD to reflect the fact that the noise level in the data gets reduced by the window, but given that the optimal SNR is independent of the data, would it make more sense as a convention to report the optimal SNR with respect to the user-supplied PSD without scaling?

ColmTalbot:
This PR fixes a longstanding bug in the likelihood when using windowed time-domain data. The addition of the window factor was due to a misunderstanding of what quantity should be whitened. The change made here will ensure that the data containing the signal is correctly whitened.

Closes #869

### DIFF
--- a/bilby/gw/detector/interferometer.py
+++ b/bilby/gw/detector/interferometer.py
@@ -492,6 +492,17 @@ class Interferometer(object):
             logger.info('  {} = {}'.format(key, parameters[key]))
 
     @property
+    def _window_power_correction(self):
+        """
+        This property enables the old (incorrect) PSD correction to be applied
+        using the :code:`BILBY_INCORRECT_PSD_NORMALIZATION` environment variable.
+        """
+        if os.environ.get("BILBY_INCORRECT_PSD_NORMALIZATION", "FALSE").upper() == "TRUE":
+            return self.strain_data.window_factor
+        else:
+            return 1
+
+    @property
     def amplitude_spectral_density_array(self):
         """ Returns the amplitude spectral density (ASD) given we know a power spectral density (PSD)
 
@@ -502,7 +513,7 @@ class Interferometer(object):
         """
         return self.power_spectral_density.get_amplitude_spectral_density_array(
             frequency_array=self.strain_data.frequency_array
-        )
+        ) * self._window_power_correction**0.5
 
     @property
     def power_spectral_density_array(self):
@@ -517,7 +528,7 @@ class Interferometer(object):
         """
         return self.power_spectral_density.get_power_spectral_density_array(
             frequency_array=self.strain_data.frequency_array
-        )
+        ) * self._window_power_correction
 
     def unit_vector_along_arm(self, arm):
         logger.warning("This method has been moved and will be removed in the future."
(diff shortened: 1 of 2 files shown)


bilby/gw/detector/interferometer.py
```diff
@@ -500,10 +500,9 @@ def amplitude_spectral_density_array(self):
         array_like: An array representation of the ASD
 
         """
-        return (
```
adivijaykumar, line 503: Do we want to introduce a option that can loosely make things backward compatible?

ColmTalbot: I considered this, we could leave it in with an environment variable flag, I don't really want it to be an option that can be toggled in the API as the current behaviour is very much a bug. I guess this could be useful as a cross check, but I don't have a specific use case. I'd be happy to hear any though.

mj-will:
> Somewhat alarmingly, this didn't require changing any of the tests...

I'd in favour of adding some then, this seems to me like something we should be able to test.

bilby/gw/detector/interferometer.py
adivijaykumar: I thought a bit about this. I think the env variable option sounds good, and I agree that toggling it in the API isn't really the best path. I was also only thinking about cross-checks with past published results, and do not have any other use case.

ColmTalbot: I've added it in, I guess an immediate use case is reweighting buggy results!

bilby/gw/detector/interferometer.py
```diff
@@ -491,6 +491,17 @@ def inject_signal_from_waveform_polarizations(self, parameters, injection_polari
         for key in parameters:
             logger.info('  {} = {}'.format(key, parameters[key]))
 
+    @property
+    def _window_power_correction(self):
+        """
+        This property enables the old (incorrect) PSD correction to be applied
+        using the :code:`BILBY_INCORRECT_PSD_NORMALIZATION` environment variable.
+        """
+        if os.environ.get("BILBY_INCORRECT_PSD_NORMALIZATION", "FALSE").upper() == "TRUE":
```
mj-will: I'd suggest using [`string_to_boolean`](https://github.com/bilby-dev/bilby/blob/main/bilby/core/utils/env.py#L3) here, it means something like `1` would work as well and is consistent with what we do else where.

mj-will (commented):
LGTM, just a comment on the environment variable.

mj-will (approved):

adivijaykumar (approved):

PR #980, opened 2025-08-21
DEP: deprecate dnest4 interface

ColmTalbot:
This PR starts the process of deprecating the native `DNest4` interface. The interface has never been widely used, the upstream package isn't actively maintained and we are seeing issues with numpy>=2 compatibility.

This PR:
- disables testing
- removes the package from our testing containers
- removes it from sampler dependencies
- adds a deprecation warning


docs/samplers.txt
```diff
@@ -72,6 +70,17 @@ MCMC samplers
...
+Deprecated samplers
+-------------------
+
+For some samplers that are not actively maintained, the maintenance burden
+outweighed the benefit of shipping the interfaces with :code:`bilby`.
+The following samplers are no longer actively supported in the main package,
+but are available via plugin packages.
+
+- PyPolyChord :code:`pypolychord-bilby`
+- DNest4 :code:`dnest4-bilby`
```
mj-will: I think should be slightly reworded since some samplers (just nessai for now) have support via plugins and that is the preferred way to use the sampler.

ColmTalbot: I moved this around somewhat.

docs/samplers.txt
```diff
@@ -53,14 +53,21 @@ recommended for direct use by the user, rather it should be accessed via the
...
 - UltraNest :code:`bilby.core.sampler.ultranest.Ultranest`
-- DNest4 :code:`bilby.core.sampler.dnest4.DNest4`
-- Nessai :code:`bilby.core.sampler.nessai.Nessai`
+
+External
+========
+
+- DNest4 :code:`dnest4-bilby`
+- Nessai :code:`nessai-bilby`
+- PyPolyChord :code:`pypolychord-bilby`
```
mj-will, line 70: I realized this is partially duplicates what is here: https://bilby-dev.github.io/bilby/plugins.html#sampler-plugin-library.

It probably makes more sense to have the list where you've added it rather than in the plugins. What do you think?

ColmTalbot: I didn't realise this information was there. I'm happy to take it out of the plugins page.

mj-will: Sounds good

mj-will (approved):
I think the samplers page could use some more details, but we can leave that for another PR.

PR #990, opened 2025-09-26
Make examples on website reproducible

fgittins:
I noticed that two examples on the website ([here](https://bilby-dev.github.io/bilby/basics-of-parameter-estimation.html) and [here](https://bilby-dev.github.io/bilby/compact-binary-coalescence-parameter-estimation.html)) were not seeded such that they would produce consistent results. I fixed this by adding an `rstate` kwarg to the `run_sampler` function.


fgittins:
Thanks for pointing that out @ColmTalbot! I must have made the changes on the wrong branch! Hopefully, it should be fixed now.

fgittins:
I've just updated the plots for the linear-regression example so that they follow from the seeding. I plan to make the plots for the gravitational-wave example too.

mj-will (approved):

PR #994, opened 2025-10-13
BUG: Update sampler_init_kwargs for dynestyv3 compatibility

ColmTalbot:
Adjust sampler initialization arguments for new dynesty API. This will fix the failures seen in https://github.com/bilby-dev/bilby/actions/runs/18467619522/job/52613334363?pr=940.

### DIFF
--- a/bilby/core/sampler/dynesty.py
+++ b/bilby/core/sampler/dynesty.py
@@ -296,7 +296,7 @@ class Dynesty(NestedSampler):
             elif kwargs["bound"] == "live":
                 logger.info(
                     "Live-point based bound method requested with dynesty sample "
-                    f"'{sample}', overwriting to 'multi'"
+                    f"'{kwargs['sample']}', overwriting to 'multi'"
                 )
                 kwargs["bound"] = "multi"
         return kwargs


bilby/core/sampler/dynesty.py
```diff
@@ -290,6 +293,8 @@ def sampler_init_kwargs(self):
             elif kwargs["sample"] == "rwalk":
                 kwargs["sample"] = dynesty_utils.AcceptanceTrackingRWalk(**kwargs)
                 kwargs["bound"] = "none"
+            elif kwargs["bound"] == "live":
+                kwargs["bound"] = "multi"
```
mj-will, line 301: Why is the check [here](https://github.com/ColmTalbot/bilby/blob/7a2581c1f9beca2d7c8d1f4cfca688ae7185b910/bilby/core/sampler/dynesty.py#L497) catching this?

ColmTalbot: Because this [whole method is skipped with the new API](https://github.com/ColmTalbot/bilby/blob/7a2581c1f9beca2d7c8d1f4cfca688ae7185b910/bilby/core/sampler/dynesty.py#L487C9-L488C19). I guess I can just move that initial check to also run for the new API. That might fit better with the existing method. What do you think?

mj-will: Ah, I missed that. In that case, I think we leave it as is since it will make removing support for the old API easier.

ColmTalbot: I guess the counterargument is that we can drop the whole other method after we stop supporting the older version of dynesty if we make the change here.

mj-will: Seems we had the same thought, shall we keep it as in then?

ColmTalbot: I think so, good to have the reminder of why I did it this way.

ColmTalbot: I copied the logging message up to here so we don't lose the useful warning.

bilby/core/sampler/dynesty.py
```diff
@@ -290,6 +293,12 @@ def sampler_init_kwargs(self):
             elif kwargs["sample"] == "rwalk":
                 kwargs["sample"] = dynesty_utils.AcceptanceTrackingRWalk(**kwargs)
                 kwargs["bound"] = "none"
+            elif kwargs["bound"] == "live":
+                logger.info(
+                    "Live-point based bound method requested with dynesty sample "
+                    f"'{sample}', overwriting to 'multi'"
```
ColmTalbot:
```suggestion
                    f"'{kwargs['sample']}', overwriting to 'multi'"
```

PR #1000, opened 2025-10-16
docs(gh-993): fix description of formula for `DirichletElement`

### LINKED ISSUE #993: incorrect formula in `bilby.core.prior.conditional.DirichletElement` docstring
I have noticed that the mentioned formula does not match the implementation,

https://github.com/bilby-dev/bilby/blob/768e1e049fde7654d457dc36364a7e632c6ed28e/bilby/core/prior/conditional.py#L332

$$
p(x_n) \propto (x_\max - x_n)^{N - n - 2}
$$

where $N$ is the dimension, $n$ is the order.

The implementation is,

$$
x_n\mid x_\max\sim B(1, N-n-1)
$$

where, $\mathrm{B}$ is the beta distribution with the $\mathrm{supp}(B)=[0,1-x_\max]$.

Simplifying the above formula gives,

$$
\forall n < N-1,\qquad p(x_n\mid x_\max) = \frac{(N - n - 1)(1-x_\max - x_n)^{N - n - 2}}{(1-x_\max)^{N - n - 1}},\qquad x_n\in[0,1-x_\max]
$$

and for $n=N-1$,

$$
x_{N-1}\mid x_\max \sim \mathrm{Unif}(0,1-x_\max)
$$

---

If you find my claim true, I will be happy to make a PR!

Qazalbash:
closes #993

### DIFF
--- a/bilby/core/prior/conditional.py
+++ b/bilby/core/prior/conditional.py
@@ -328,6 +328,7 @@ class DirichletElement(ConditionalBeta):
 
     .. math::
         p(x_n\mid S_n) := \begin{cases}
+            1 & N = 1 \\
             \displaystyle\frac{(N - n - 1)(1 - S_n - x_n)^{N - n - 2}}{(1-S_n)^{N-n-1}}
             & n < N - 1 \\
             \displaystyle\delta(1-S_{N-1}-x_{N-1}) & n = N - 1
@@ -337,7 +338,7 @@ class DirichletElement(ConditionalBeta):
     dimensions and :math:`S_n` is the sum of all previous dimensions,
 
     .. math::
-        S_n := \sum_{i=0}^{n-1}x_i & n < N
+        S_n := \sum_{i=0}^{n-1}x_i \qquad n < N
 
     Examples
     ========
@@ -350,10 +351,10 @@ class DirichletElement(ConditionalBeta):
         :code:`n_dimensions` = :math:`N = 2`:
 
             .. math::
-                \begin{align}
-                    p(x_0)         &= 1                ; & 0 \leq x_0 \leq 1       \\
-                    p(x_1\mid x_0) &= \delta(1-x_0-x_1); & 0 \leq x_1 \leq 1 - x_0
-                \end{align}
+                    p(x_0)         = 1                ; \qquad 0 \leq x_0 \leq 1
+
+            .. math::
+                    p(x_1\mid x_0) = \delta(1-x_0-x_1); \qquad 0 \leq x_1 \leq 1 - x_0
 
     Parameters
     ==========


bilby/core/prior/conditional.py
```diff
@@ -323,39 +323,48 @@ class ConditionalInterped(conditional_prior_factory(Interped)):
 
 
 class DirichletElement(ConditionalBeta):
-    r"""
-    Single element in a dirichlet distribution
-
-    The probability scales as
+    r"""Single element in a dirichlet distribution. The probability defined as,
```
ColmTalbot: Multiline docstrings should start with a blank line.


```suggestion
    r"""
    Single element in a dirichlet distribution. The probability defined as,
```

bilby/core/prior/conditional.py
```diff
@@ -323,39 +323,48 @@ class ConditionalInterped(conditional_prior_factory(Interped)):
...
 
     .. math::
-        p(x_n) \propto (x_\max - x_n)^{(N - n - 2)}
+        p(x_n\mid S_n) := \begin{cases}
+			\displaystyle\frac{(N - n - 1)(1 - S_n - x_n)^{N - n - 2}}{(1-S_n)^{N-n-1}}
+            & n < N - 1 \\
+			\displaystyle\frac{1}{1-S_n} & n = N - 1
```
ColmTalbot: I don't think this is correct, the $n = N - 1$ term is not formally implemented by this class but should be set to $1 - S_{N - 1}$, i.e., it is a delta function.

bilby/core/prior/conditional.py
ColmTalbot: Multiline comments should start with a line break.


```suggestion
    r"""
    Single element in a dirichlet distribution. The probability defined as,
```

bilby/core/prior/conditional.py
Qazalbash: Fixed!

bilby/core/prior/conditional.py
```diff
@@ -324,38 +324,49 @@ class ConditionalInterped(conditional_prior_factory(Interped)):
...
+            \displaystyle\delta(1-S_{N-1}-x_{N-1}) & n = N - 1
+        \end{cases}, \qquad 0 \leq x_n \leq 1 - S_n
+
+    where, :math:`\delta` is the Dirac delta function, :math:`N` is the total number of
+    dimensions and :math:`S_n` is the sum of all previous dimensions,
 
-    for :math:`x_n < x_\max`, where :math:`x_\max` is the sum of :math:`x_i`
-    for :math:`i < n`
+    .. math::
+        S_n := \begin{cases}
```
ColmTalbot: This can be simplified, the special case for $n = N - 1$ doesn't need to be there (and is currently incorrect.)


```suggestion
        S_n = \displaystyle\sum_{i=0}^{N-2}x_i
```

bilby/core/prior/conditional.py
```diff
@@ -324,38 +324,49 @@ class ConditionalInterped(conditional_prior_factory(Interped)):
...
+                    0                                & n = N - 1 \\
```
ColmTalbot: To be clear, I'm talking about this line.

Qazalbash: Yeah, I get it.

ColmTalbot (approved):
This looks good to me now!

bilby/core/prior/conditional.py
```diff
@@ -324,38 +324,46 @@ class ConditionalInterped(conditional_prior_factory(Interped)):
...
-        .. math::
-            p(x_0) \propto 1 ; 0 < x_0 < 1
+        :code:`n_dimensions` = :math:`N = 1`:
 
-    n_dimensions = 2:
-        .. math::
-            p(x_0) &\propto (1 - x_0) ; 0 < x_0 < 1
-            p(x_1) &\propto 1 ; 0 < x_1 < 1
+            .. math::
+                p(x_0) = 1 ; 0 < x_0 < 1
```
mj-will: I might be missing something, but following the equation about I think this should be:
```
p(x_0) = \delta(1 - x_0)
```

Also, should these `\leq` rather than `<`?

bilby/core/prior/conditional.py
```diff
@@ -324,38 +324,46 @@ class ConditionalInterped(conditional_prior_factory(Interped)):
...
+
+        :code:`n_dimensions` = :math:`N = 2`:
+
+            .. math::
+                \begin{align}
+                    p(x_0)         &= 1 - x_0          ; & 0 \leq x_0 \leq 1       \\
```
mj-will: Looking at the equation again, this corresponds to the top case, so:

```
(2 - 0 - 1) (...)^(2 - 0 - 2) / (1 - 0)^(2- 0 -1)
```

which I think reduces to `1` rather than `1 - x_0`. Again, I might be missing something so let me know what you think.

Qazalbash: I think you are right. I will update it shortly.

bilby/core/prior/conditional.py
Qazalbash: I agree with the second part. For the second part, I have to look into the derivation I did.

PR #1002, opened 2025-10-16
ENH: add npool_post_process

mj-will:
Add an `npool_post_process`.

This allows the user to configure how many processes are used for post-processing independently of how many are used for sampling. Whilst making this change, I also changed the code so aliases for npool also trigger the parallelized post-processing.

I also remove an old unused argument.


mj-will:
I suggest we leave this until after https://github.com/bilby-dev/bilby/pull/1009

PR #1003, opened 2025-10-16
STY: apply ruff formatting

### LINKED ISSUE #977: MAINT: Switching to ruff formatting
As suggested in https://github.com/bilby-dev/bilby/pull/952#discussion_r2147155454, we should considering switching from black+flake8 to ruff

ColmTalbot:
It looks like copilot refused my request to work on this so here's a PR that closes #977.

This should not be squashed as I added a file to ignore the commit where everything changed when blaming (it mostly works, but gets a bit confused by blank lines).

I made some editorial choices in how we import things/define `__all__`. We should probably clean up the namespace at some stage, but I refrained from doing that here and added explicit `noqa` tags when needed to support that.


mj-will (approved):
@ColmTalbot do you want to get this in before or after the backend changes?

adivijaykumar:
It feels like it should go in after the backend changes?

PR #1009, opened 2025-10-30
FEAT: Improve interactions of user pools

ColmTalbot:
I noticed that it was difficult to pass a user-specified pool through run_sampler, and it isn't used at all in the post processing. This PR:

- unifies how to create a pool for Bilby with a context (a la `with multiprocessing.pool()...`)
- explicitly passes the user pool around including for post processing
- moves logic out of `Sampler._setup_pool` and `Sampler._close_pool`
- support MPI pools using schwimmbad. This change will effectively make `parallel_bilby` moot.

### DIFF
--- a/bilby/core/result.py
+++ b/bilby/core/result.py
@@ -247,11 +247,11 @@ def get_weights_for_reweighting(
     def eval_pool(this_logl):
         from .utils.parallel import bilby_pool
 
-        chunksize = max(100, n // (2 * npool))
         with bilby_pool(likelihood=this_logl, npool=npool) as my_pool:
             if my_pool is None:
                 map_fn = map
             else:
+                chunksize = max(100, n // (2 * npool))
                 map_fn = partial(my_pool.imap, chunksize=chunksize)
             likelihood_fn = partial(_safe_likelihood_call, this_logl)
 
--- a/bilby/core/result.py
+++ b/bilby/core/result.py
@@ -252,10 +252,11 @@ def get_weights_for_reweighting(
         if my_pool is None:
             map_fn = map
         else:
-            map_fn = my_pool.imap
+            map_fn = partial(my_pool.imap, chunksize=chunksize)
+        likelihood_fn = partial(_safe_likelihood_call, this_logl)
 
         log_l = list(tqdm(
-            map_fn(partial(_safe_likelihood_call, this_logl), dict_samples[starting_index:], chunksize=chunksize),
+            map_fn(likelihood_fn, dict_samples[starting_index:]),
             desc='Computing likelihoods',
             total=n,
         ))
--- a/bilby/core/result.py
+++ b/bilby/core/result.py
@@ -253,10 +253,9 @@ def get_weights_for_reweighting(
             else:
                 chunksize = max(100, n // (2 * npool))
                 map_fn = partial(my_pool.imap, chunksize=chunksize)
-            likelihood_fn = partial(_safe_likelihood_call, this_logl)
 
             log_l = list(tqdm(
-                map_fn(likelihood_fn, dict_samples[starting_index:]),
+                map_fn(this_logl.log_likelihood, dict_samples[starting_index:]),
                 desc='Computing likelihoods',
                 total=n,
             ))
--- a/bilby/core/result.py
+++ b/bilby/core/result.py
@@ -9,6 +9,7 @@ from copy import copy
 from importlib import import_module
 from itertools import product
 import multiprocessing
+from functools import partial
 import numpy as np
 import pandas as pd
 import scipy.stats
@@ -31,6 +32,11 @@ from .prior import Prior, PriorDict, DeltaFunction, ConditionalDeltaFunction
 
 
 EXTENSIONS = ["json", "hdf5", "h5", "pickle", "pkl"]
+
+
+def __eval_l(likelihood, params):
+    likelihood.parameters.update(params)
+    return likelihood.log_likelihood()
 
 
 def result_file_name(outdir, label, extension='json', gzip=False):
@@ -191,7 +197,7 @@ def read_in_result_list(filename_list, invalid="warning"):
 
 def get_weights_for_reweighting(
         result, new_likelihood=None, new_prior=None, old_likelihood=None,
-        old_prior=None, resume_file=None, n_checkpoint=5000, npool=1):
+        old_prior=None, resume_file=None, n_checkpoint=5000, npool=1, pool=None):
     """ Calculate the weights for reweight()
 
     See bilby.core.result.reweight() for help with the inputs
@@ -234,8 +240,7 @@ def get_weights_for_reweighting(
         basedir = os.path.split(resume_file)[0]
         check_directory_exists_and_if_not_mkdir(basedir)
 
-    dict_samples = [{key: sample[key] for key in result.posterior}
-                    for _, sample in result.posterior.iterrows()]
+    dict_samples = result.posterior.to_dict(orient="records")
     n = len(dict_samples) - starting_index
 
     # Helper function to compute likelihoods in parallel
@@ -243,11 +248,8 @@ def get_weights_for_reweighting(
         with multiprocessing.Pool(processes=npool) as pool:
             chunksize = max(100, n // (2 * npool))
             return list(tqdm(
-                pool.imap(
-                    this_logl.log_likelihood,
-                    dict_samples[starting_index:],
-                    chunksize=chunksize,
-                ),
+                pool.imap(partial(__eval_l, this_logl),
+                        dict_samples[starting_index:], chunksize=chunksize),
                 desc='Computing likelihoods',
                 total=n)
             )
@@ -321,7 +323,7 @@ def rejection_sample(posterior, weights):
 def reweight(result, label=None, new_likelihood=None, new_prior=None,
              old_likelihood=None, old_prior=None, conversion_function=None, npool=1,
              verbose_output=False, resume_file=None, n_checkpoint=5000,
-             use_nested_samples=False):
+             use_nested_samples=False, pool=None):
     """ Reweight a result to a new likelihood/prior using rejection sampling
 
     Parameters
@@ -384,7 +386,9 @@ def reweight(result, label=None, new_likelihood=None, new_prior=None,
         get_weights_for_reweighting(
             result, new_likelihood=new_likelihood, new_prior=new_prior,
             old_likelihood=old_likelihood, old_prior=old_prior,
-            resume_file=resume_file, n_checkpoint=n_checkpoint, npool=npool)
+            resume_file=resume_file, n_checkpoint=n_checkpoint,
+            npool=npool, pool=pool,
+        )
 
     if use_nested_samples:
         ln_weights += np.log(result.posterior["weights"])
@@ -411,10 +415,14 @@ def reweight(result, label=None, new_likelihood=None, new_prior=None,
 
     if conversion_function is not None:
         data_frame = result.posterior
-        if "npool" in inspect.signature(conversion_function).parameters:
-            data_frame = conversion_function(data_frame, new_likelihood, new_prior, npool=npool)
-        else:
-            data_frame = conversion_function(data_frame, new_likelihood, new_prior)
+        parameters = inspect.signature(conversion_function).parameters
+        kwargs = dict()
+        for key, value in [
+            ("likelihood", new_likelihood), ("priors", new_prior), ("npool", npool), ("pool", pool)
+        ]:
+            if key in parameters:
+                kwargs[key] = value
+        data_frame = conversion_function(data_frame, **kwargs)
         result.posterior = data_frame
 
     if label:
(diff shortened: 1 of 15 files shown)


bilby/core/sampler/__init__.py
```diff
@@ -305,42 +298,81 @@ def run_sampler(
     if sampler.cached_result:
         logger.warning("Using cached result")
         result = sampler.cached_result
+        result = apply_conversion_function(
+            result=result,
+            likelihood=likelihood,
+            conversion_function=conversion_function,
+            npool=npool,
+            pool=pool,
+        )
```
mj-will, line 320: Will this always reapply the conversion function? If so, I'm unsure if this is desirable since rather than bilby quickly exiting when a run is already done it will spend time doing the conversion. That said, i don't feel strongly about this, so happy to keep it. Thoughts?

bilby/core/sampler/__init__.py
```diff
@@ -305,42 +298,81 @@ def run_sampler(
...
+            likelihood,
+            priors,
+            use_ratio=sampler.use_ratio,
+            search_parameter_keys=sampler.search_parameter_keys,
+            npool=npool,
+            pool=pool,
+            parameters=priors.sample(),
+        ) as _pool:
+            start_time = datetime.datetime.now()
+            sampler.pool = _pool
```
mj-will, line 333: Is this safe to do? Depending on how the sampler uses the pool, are there settings where the pool the sampler has stored is not updated?

For example, if the sampler has constructed a likelihood using the `pool.map` from the initial input `pool` will this break?

bilby/core/sampler/ptemcee.py
```diff
@@ -417,7 +417,7 @@ def setup_sampler(self):
 
         if (
             os.path.isfile(self.resume_file)
-            and os.path.getsize(self.resume_file)
+            and os.path.getsize(self.resume_file) > 5
```
mj-will: Why is this needed?

bilby/core/utils/parallel.py
```diff
@@ -0,0 +1,87 @@
...
+
+@contextmanager
+def bilby_pool(
+    likelihood, priors,
+    use_ratio=None,
+    search_parameter_keys=None,
+    npool=None,
+    pool=None,
+    parameters=None,
+):
```
mj-will, line 202: I think this would benefit from a doc-string with example usage.

bilby/core/utils/parallel.py
```diff
@@ -0,0 +1,87 @@
...
+
+def create_pool(
+    likelihood=None,
+    priors=None,
+    use_ratio=None,
+    search_parameter_keys=None,
+    npool=None,
+    pool=None,
+    parameters=None,
+):
```
mj-will, line 76: This would benefit from a doc-string. In particular, I think we should make clear what happens in `npool` is None or 1.

bilby/core/result.py
```diff
@@ -235,20 +235,28 @@ def get_weights_for_reweighting(
...
+        else:
+            map_fn = partial(my_pool.imap, chunksize=chunksize)
+        likelihood_fn = partial(_safe_likelihood_call, this_logl)
+
+        log_l = list(tqdm(
+            map_fn(likelihood_fn, dict_samples[starting_index:]),
+            desc='Computing likelihoods',
+            total=n,
+        ))
+        close_pool(my_pool)
```
mj-will: Any reason to use open/close over the context manager?

bilby/gw/conversion.py
```diff
@@ -2329,49 +2319,35 @@ def compute_per_detector_log_likelihoods(samples, likelihood, npool=1, block=10)
...
+                        _compute_per_detector_log_likelihoods,
+                        fill_args[ii: ii + block]
+                    )
+                else:
+                    from ..core.sampler.base_sampler import _sampling_convenience_dump
+                    _sampling_convenience_dump.likelihood = likelihood
+                    subset_samples = [
+                        list(_compute_per_detector_log_likelihoods(xx))
+                        for xx in fill_args[ii: ii + block]
+                    ]
```
mj-will: Is it worth simplifying this with something like this?
```python
if _pool is None:
    map_fn = map
    _sampling_convenience_dump.likelihood = likelihood
else:
    map_fn = pool.map

 subset_samples = map_fn(
                        _compute_per_detector_log_likelihoods,
                        fill_args[ii: ii + block]
                    )
```

bilby/gw/conversion.py
```diff
@@ -2472,51 +2448,31 @@ def generate_posterior_samples_from_marginalized_likelihood(
...
+                continue
 
-        if pool is not None:
-            subset_samples = pool.map(fill_sample, fill_args[ii: ii + block])
-        else:
-            subset_samples = [list(fill_sample(xx)) for xx in fill_args[ii: ii + block]]
+            if _pool is not None:
+                subset_samples = _pool.map(fill_sample, fill_args[ii: ii + block])
+            else:
+                subset_samples = [list(fill_sample(xx)) for xx in fill_args[ii: ii + block]]
```
mj-will: Same comment about `map` but not sure it's worth it. Thoughts?

mj-will (commented):
Looks great, I look forward to trying it! I've added some initial comments but might need to have a second look.

bilby/core/sampler/ptemcee.py
ColmTalbot: Empty pickle files have nonzero size (5B) so currently the code thinks that should be a valid resume file and fails. Probably this should be fixed on main too.

bilby/core/sampler/__init__.py
ColmTalbot: It will reapply it, this maintains the current behaviour. I'd be open to changing that behaviour, maybe by adding a flag to the result file to say if the conversion has been applied, but I would say to do that as a separate change.

bilby/core/sampler/__init__.py
ColmTalbot: I'm trying to work through what this would look like.

If the initial input `pool` is not None, all of the pool objects reference here should be the same, so I don't think that will make a difference.

One potential issue is that if a specific sampler implementation handles the pool internally by itself this will create two pools and then in the best case, we have a bunch of extra processes we don't need. I think this is what nessai does, so maybe we should game through that specific case.

ColmTalbot (commented):
I don't remember exactly what the behaviour for some of these questions is, so I'll go back and check, and ideally write docstrings about them.

bilby/core/result.py
ColmTalbot: It seems not.

bilby/gw/conversion.py
ColmTalbot: Sure

bilby/gw/conversion.py
ColmTalbot: I modified it to look closer, but I agree it probably isn't worth it in this case.

bilby/core/sampler/__init__.py
mj-will: Coming back to this, would it make sense to have an `update_pool` method or similar? That could have some additional logic to prevent multiple pools etc.

bilby/core/result.py
mj-will: I suggest switching to a context manager then since I think it's more robust to exceptions.

mj-will (changes requested):
Forgot to submit my comments as a review, but see them above.

bilby/core/utils/parallel.py
```diff
@@ -0,0 +1,204 @@
...
+        if _pool.is_master():
+            logger.info(f"Created MPI pool with size {_pool.size}")
+    elif pool is not None:
+        _pool = pool
+    elif npool not in (None, 1):
+        import multiprocessing
+
+        _pool = multiprocessing.Pool(
+            processes=npool,
+            initializer=_initialize_global_variables,
```
mj-will: I think it would be useful to allow users to specify their own (additional) initializer.

For nessai, there's another variable that needs to be global for the pool to work and that can't be intialized unless using the `fork` start method.

ColmTalbot: Does this mean that we would also need to be able to allow users to pass through arbitrary additional arguments. At that stage, is it easier to just describe how to implement an independent version?

bilby/core/result.py
ColmTalbot: It looks like I changed this at some stage.

bilby/core/utils/parallel.py
mj-will: Yes, that's true.

Here's what I had to do for `nessai`: 

- https://github.com/bilby-dev/nessai-bilby/blob/main/src/nessai_bilby/plugin.py#L29
- https://github.com/bilby-dev/nessai-bilby/blob/main/src/nessai_bilby/plugin.py#L362

It works fine but since it duplicates so much of the code, it needs to be kept up-to-date with the main code base.

bilby/core/utils/parallel.py
```diff
@@ -0,0 +1,204 @@
...
+    >>>     likelihood = _sampling_convenience_dump.likelihood
+    >>>     return likelihood.log_likelihood(parameters)
+
+    >>> likelihood = AnalyticalMultidimensionalCovariantGaussian(
+    >>>     mean=np.zeros(4), cov=np.eye(4)
+    >>> )
+    >>> priors = PriorDict({f"x{i}": Normal(0, 1) for ii in range(4)})
+    >>> parameters = [priors.sample() for _ in range(10)]
+    >>> with bilby_pool(likelihood, priors, npool=4) as pool:
+    >>>     log_ls = list(pool.map(some_parallel_function, parameters))
```
ColmTalbot:
```suggestion
    >>>     log_ls = list(pool.map(parallel_likelihood_eval, parameters))
```

adivijaykumar (commented):
Can we add a test for schwimmbad/MPI support?

mj-will:
We should documentation for user pools and note that the pool will only work correctly in the initializer has been called.

bilby/core/sampler/__init__.py
mj-will: Here's an example of what I had in mind: https://github.com/mj-will/aspire-bilby/blob/main/src/aspire_bilby/plugin.py#L199

`self.pool` is used in to create the likelihood function for `aspire`.

Looking at this now, since it happens in `run_sampler` perhaps this isn't actually an issue in this specific case. What do you think?

ColmTalbot: Yeah, since this is inside the `Sampler.run_sampler` method this should be fine.

bilby/core/utils/parallel.py
mj-will: We've agreed to leave this for a future change since this can be fixed per-sampler and improved later.

mj-will (changes requested):
Coming back to this again, I think the only pending item is documentation for the new pool support.

bilby/core/sampler/base_sampler.py
```diff
@@ -768,41 +730,44 @@ def _raise_if_interrupted(self, cause):
...
+                use_ratio=self.use_ratio,
+                npool=self.npool,
+                pool=self.pool,
+                parameters=parameters,
+            )
+            if self.pool is not None:
+                logger.warning(
+                    "Setting up parallel pool in sampler is deprecated. Use "
+                    "bilby.utils.parallel.bilby_pool context instead."
+                )
```
mj-will, line 764: Does this mean we need to change how the pool is called in bilby_pipe?

docs/parallelisation.rst
```diff
@@ -0,0 +1,140 @@
...
+each worker process. We support multiple ways of performing this initialization.
+
+.. note::
+
+    Recent versions of Python include freethreaded builds allowing efficient
+    thread-based parallelism. We do not currently support thread-based parallelism
+    due to the use of global storage of the likelihood and priors.
+    However, we plan to remove this limitation in a future release.
+
+Default parallelisation
```
mj-will, line 24: I'm a bit confused by this given there's a deprecation warning above. I think I'm missing somethign about the deprecation but I'm not sure what.

mj-will (commented):
Overall LGTM, just one question on the deprecation.

bilby/core/sampler/base_sampler.py
ColmTalbot: I can probably clarify this, as the call order is pretty confusing. This should only be triggered if people aren't using `bilby.run_sampler`, which is a pretty minimal subset of people who are directly invoking the sampler classes. In that case, the recommendation is to use the bilby_pool context.

If you have a suggestion for more clear text, I'd be open to it.

mj-will: Ah, that makes sense.

Perhaps clarifying which method is being called would help? Maybe something like:


```suggestion
            if self.pool is not None:
                logger.warning(
                    "Setting up parallel pool in Sampler._setup_pool is deprecated. "
                    "Use bilby.utils.parallel.bilby_pool context instead."
                )
```

docs/parallelisation.rst
ColmTalbot: Is this addressed with the change to the warning?

mj-will: Yes, thanks

mj-will (approved):
Thanks for addressing those points Colm, LGTM now.

adivijaykumar (approved):
Looks good to me and happy to approve! Out of curiosity @ColmTalbot have you benchmarked the schwimmbad/MPI functionality? If yes, I think it would be nice to advertise on a PE call (or some such venue) that people can use this instead of parallel_bilby.

PR #1014, opened 2025-11-03
CI: Modify publish.yml to support TestPyPI

### LINKED ISSUE #978: Use test.pypi.org for testing packaging changes
Following https://github.com/bilby-dev/bilby/pull/952, we should consider testing the build process uses test.pypi.org

ColmTalbot:
Updated the GitHub Actions workflow to trigger on push events and added a new job for publishing to TestPyPI.

Closes #978

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,6 +2,8 @@ name: Publish Python distribution to PyPI
 
 on:
   - push
+    branches:
+      - main
   - release
 
 jobs:


.github/workflows/publish.yml
```diff
@@ -1,7 +1,6 @@
 name: Publish Python distribution to PyPI
-on:
-  release:
-    types: [published]
+
+on: push
```
ColmTalbot: Workflow needs to be triggered on release.

.github/workflows/publish.yml
```diff
@@ -1,7 +1,8 @@
 name: Publish Python distribution to PyPI
+
 on:
-  release:
-    types: [published]
+  - push
```
mj-will, line 4: Should we specify the main branch here? With out, I think that means pushes to any PR will make a test release.

mj-will (commented):
Just one question

.github/workflows/publish.yml
ColmTalbot:
```suggestion
  - push
    branches:
      - main
```

GregoryAshton (approved):

mj-will (approved):

PR #1016, opened 2025-11-05
BUG: fix triangular interferometer coordinates

filippo-santoliquido:
Hi,

I believe I have found two bugs in the implementation of the triangular-shaped interferometer.
The first issue concerns [this test](https://github.com/bilby-dev/bilby/blob/main/test/gw/detector/networks_test.py#L361), where the angles passed to [this function](https://github.com/bilby-dev/bilby/blob/main/test/gw/detector/networks_test.py#L390) should be given in radians, as also stated in [1]. However, in the current implementation, the last two latitude values are passed in degrees.
When the angles are passed correctly, the test fails. For example, if the position of the first vertex is set to `(40.516°, 9.416°)` with `x_arm_azimuth = 70.5674°`, the distances between the vertices are:

ET-1 ↔️ ET-2: 9.06 km
ET-1 ↔️ ET-3: 9.76 km
ET-2 ↔️ ET-3: 7.70 km

I believe this test failure is due to an incorrect implementation of the functions updating [latitude](https://github.com/bilby-dev/bilby/blob/main/bilby/gw/detector/networks.py#L382) and [longitude](https://github.com/bilby-dev/bilby/blob/main/bilby/gw/detector/networks.py#L392), which differ from those provided in [1]:

```
const φ2 = Math.asin( Math.sin(φ1)*Math.cos(d/R) +
Math.cos(φ1)*Math.sin(d/R)*Math.cos(brng) );
const λ2 = λ1 + Math.atan2(Math.sin(brng)*Math.sin(d/R)*Math.cos(φ1),
Math.cos(d/R)-Math.sin(φ1)*Math.sin(φ2));
```

The only difference to keep in mind is the definition of the bearing angle: in bilby it is defined counterclockwise from east, whereas in [1] it is clockwise from north. This leads to the following conversion:

```
brng = 90 - x_arm_azimuth
```

Using the above expressions, the coordinates are recovered correctly.
I also include a map illustrating this result: the blue dots represent the incorrect coordinates, while the red dots show the corrected ones.

ET_map

I believe this also resolves [this issue](https://git.ligo.org/lscsoft/bilby/-/issues/655), since, when using the Virgo site coordinates as the starting point, it becomes possible to recover the positions of the other two vertices correctly, as defined in [LAL](https://lscsoft.docs.ligo.org/lalsuite/lal/_l_a_l_detectors_8h_source.html#l00602).

I don’t think this bug has any practical effect on previously obtained parameter estimation results.

CheckFix_log_likelihood

I’m opening this pull request to contribute to fixing this bug.

I am tagging a few people who might be interested in the matter [@jacopo.tissino](https://github.com/jacopo.tissino) [@sylvia.biscoveanu](https://github.com/sylvia.biscoveanu)

Thanks a lot in advance.
Kind regards,
Filippo

[1] [https://www.movable-type.co.uk/scripts/latlong.html](https://www.movable-type.co.uk/scripts/latlong.html)

### DIFF
--- a/bilby/gw/detector/networks.py
+++ b/bilby/gw/detector/networks.py
@@ -358,6 +358,8 @@ class TriangularInterferometer(InterferometerList):
         if isinstance(maximum_frequency, float) or isinstance(maximum_frequency, int):
             maximum_frequency = [maximum_frequency] * 3
 
+        brng = 90 - xarm_azimuth
+
         for ii in range(3):
             self.append(
                 Interferometer(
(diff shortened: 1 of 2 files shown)


filippo-santoliquido:
Hi Colm,

Thanks for pointing me to #894 

The coordinates produced by the two methods are essentially identical
```
#1016 : (0.7615118398425819, 0.18333805212847526), (0.7629902506868369, 0.18405974788582488), (0.7627015830984775, 0.18192790169194423)
```

```
#894 : (0.7615118398425819, 0.18333805212847526) (0.7629930799428372, 0.1840585887020974), (0.762704632610569, 0.18192996730123293)
```

though the method implemented in #894  yields slightly more precise results, as it also updates the elevation. This level of precision is, of course, important for the null stream.

At this point, I think the choice of which PR to accept is up to you. My PR applies minimal changes to the current implementation, while #894  version is more precise and may be useful for specific studies.

In any case, the [testing bug](https://github.com/bilby-dev/bilby/pull/1016/files#diff-63cb0b6a2cf41bc181313e66ff253f834c1136f9908af39ffa2289a1232c1d4d) should be fixed regardless.

Cheers,
Filippo

mj-will (approved):

PR #1022, opened 2025-12-09
BUG: Fix handling of contour-kwargs for multiple corner plots

JasperMartins:
Hi everyone,

At the moment, it is not possible to pass `contour_kwargs` to the `plot_multiple` corner-plot function because `plot_multiple` passes its own `contour_kwargs` to `result.plot_corner` even if it was provided as a keyword argument.
This little PR fixes this. The PR also removes an intermediate plotting call, which was made redundant by https://github.com/bilby-dev/bilby/pull/946.

### DIFF
--- a/bilby/core/result.py
+++ b/bilby/core/result.py
@@ -2111,7 +2111,7 @@ def plot_multiple(results, filename=None, labels=None, colours=None,
         `corner_labels` input).
         However, `show_titles` and `truths` are ignored since they would be
         ambiguous on such a plot. The keyword arguments `contour_kwargs["linestyles"]`,
-        `contour_kwargs['colors']`, `hist_kwargs["linestyle"]`, `color` and 
+        `contour_kwargs['colors']`, `hist_kwargs["linestyle"]`, `color` and
         `hist_kwargs["color"]` are overwritten with the values provided in the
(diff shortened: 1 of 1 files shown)


JasperMartins:
Good call - actually revealed a small bug with the color setting I had not noticed before.

Old output of the plot_multiple test:
multiple_old
New output of the test:
multiple_new

New output if passing a contour_kwargs dict setting linewidhts zero:
multiple_new_no_lines

bilby/core/result.py
```diff
@@ -2110,7 +2110,10 @@ def plot_multiple(results, filename=None, labels=None, colours=None,
         for the keyword `labels` for which you should use the dedicated
         `corner_labels` input).
         However, `show_titles` and `truths` are ignored since they would be
-        ambiguous on such a plot.
+        ambiguous on such a plot. The keyword arguments `contour_kwargs["linestyles"]`,
+        `hist_kwargs["linestyle"]`, `color` and `hist_kwargs["color"]` are overwritten
+        with the values provided in the `colours` and `linestyles` inputs or by the default
+        styles.
```
mj-will: Given the latest change, should this also mention the `colors` arguments in `contour_kwargs`?

PR #1023, opened 2025-12-12
[BUG] Fix the handling of conditional priors dependent on JointPrior in ConditionalPriorDict

JasperMartins:
This PR addresses the issues in `ConditionalPriorDict.rescale` described in the first item  in https://github.com/bilby-dev/bilby/issues/1026

Just to reiterate, the addressed issues are:
 - Currently, if any prior in `ConditionalPriorDict` depends on a `JointPrior`, the correct ordering of the keys is not ensured for rescaling because all keys associated with the `JointPrior` need to be rescaled first.
 - Even if the order of keys was correct, rescaling would fail because the `least_recently_sampled` property of the joint priors is not updated before it is accessed by `get_required_variables`.
 

Note: The `safe_flatten` function introduced in https://github.com/bilby-dev/bilby/pull/979 to `ConditionalPriorDict.rescale` has a bug that if any of the values was an array, the returned flattened array takes `result[key].flatten()` where it should use `value.flatten()`. However, I believe the flattening operation is redundant now anyway. It was previously only used to flatten the output of `JointPriors`, I believe, which is now handled separately, and the flattening is never triggered by any of the test cases. (Note that if one attempts to rescale many samples at once, which works in many cases and is explicitly supported by some priors' doc-strings, this bug would lead to a wrong output. However, flattening would also not be necessary here.)

### DIFF
--- a/bilby/core/prior/dict.py
+++ b/bilby/core/prior/dict.py
@@ -885,14 +885,16 @@ class ConditionalPriorDict(PriorDict):
             if isinstance(self[key], JointPrior):
                 # if joint prior, keep track if all names have been rescaled
                 distname = self[key].dist.distname
-                names = set(self[key].dist.names)
+                # maintain order of names as in the dist as this is the order
+                # in which they will be rescaled
+                names = self[key].dist.names
                 if distname not in joint:
                     joint[distname] = {key}
                 else:
                     joint[distname].add(key)
                 # only when all names have been rescaled, we can set the values
                 # we use sets because the order does not matter here
-                if names == joint[distname]:
+                if set(names) == joint[distname]:
                     for name, value in zip(names, result[key]):
                         result[name] = value
                         self[name].least_recently_sampled = value


bilby/core/prior/dict.py
```diff
@@ -871,36 +874,31 @@ def rescale(self, keys, theta):
...
-                return value
+
+            # if any requested key depends on some joint prior `jp_key`
+            # self[jp_key].least_recently_sampled needs to be set before
+            # rescaling those requested keys.
+            # Thus we keep track of joint priors here
+            if isinstance(self[key], JointPrior):
+                # if joint prior, keep track if all names have been rescaled
+                distname = self[key].dist.distname
+                names = set(self[key].dist.names)
```
ColmTalbot: This feels like it shouldn't be needed, do we not impose that parameter names are unique?

bilby/core/prior/dict.py
```diff
@@ -871,36 +874,31 @@ def rescale(self, keys, theta):
...
+                if distname not in joint:
+                    joint[distname] = [key]
```
ColmTalbot: Since this is only used as a set, why not just construct this directly as a set?

ColmTalbot (commented):
Thanks for breaking this down, I think the logic makes sense. I just had a couple of minor questions to make sure we aren't doing unnecessary conversions between lists/sets.

It's great to see the end of `safe_flatten`!

bilby/core/prior/dict.py
JasperMartins: The cast to set is made because it does not matter in which order the keys occur in `dist.names` or `joint[distname]`. This could also be solved with a loop, if that would be preferred.

bilby/core/prior/dict.py
JasperMartins: True, and there were other redundant casts. Fixed with a new commit!

JasperMartins: Note: Noticed there was one more simplification possible

JasperMartins:
There was a subtle bug in casting to `JointPrior.dist.names` too early, because the joint prior is always rescaled in the same order as `JointPrior.dist.names`. On the other hand, sometimes, `zip(set(JointPrior.dist.names, result))` has the wrong order. If this happens appears to be effectively random.

Fixed now.

PR #1029, opened 2025-12-22
BUG: Fix GWSignalWaveformGenerator reconstruction in waveform posterior plots

lorenzopompili00:
I was trying to plot some waveform reconstructions using `plot_interferometer_waveform_posterior` for some eccentric PE runs using the new `GWSignalWaveformGenerator`, and I realized it is currently not working correctly. 

The reason is that the arguments required to instantiate `GWSignalWaveformGenerator` are not saved in the likelihood metadata, and are not passed to the waveform generator when it is instantiated in `plot_interferometer_waveform_posterior` [here](https://github.com/bilby-dev/bilby/blob/main/bilby/gw/result.py#L420-L426).

A fix is to save and pass these extra arguments whenever the waveform generator is `GWSignalWaveformGenerator`. I have added this via a `waveform_generator_constructor_dict`, inspired by this `bilby_pipe` [merge request](https://git.ligo.org/lscsoft/bilby_pipe/-/merge_requests/640). Let me know if you have any other preferences for handling this.

### DIFF
--- a/test/gw/likelihood_test.py
+++ b/test/gw/likelihood_test.py
@@ -225,6 +225,7 @@ class TestGWTransient(unittest.TestCase):
             reference_frame="sky",
             lal_version=self.likelihood.lal_version,
             lalsimulation_version=self.likelihood.lalsimulation_version,
+            waveform_generator_meta_data=self.waveform_generator.meta_data,
         )
         self.assertDictEqual(expected, self.likelihood.meta_data)
 


bilby/gw/likelihood/base.py
```diff
@@ -1181,3 +1182,10 @@ def meta_data(self):
             reference_frame=self._reference_frame_str,
             lal_version=self.lal_version,
             lalsimulation_version=self.lalsimulation_version)
+
+        if isinstance(self.waveform_generator, GWSignalWaveformGenerator):
```
ColmTalbot: Maybe we should just add a generic meta_data attribute to the waveform generator as for interferometers and then add that here? That would avoid having a special case here.

mj-will:
> @lorenzopompili00 thanks for opening this PR! I think we should add this functionality, but I'd like to think about whether we can find a more generic way of implementing this.
> 
> To be concrete, I'd suggest adding a new property to the `WaveformGenerator` like
> 
> ```python
> class WaveformGenerator:
>     ...
>     @property
>     def meta_data(self):
>         return dict()
> 
> 
> class GWSignalWaveformGenerator:
>     ...
>     @property
>     def meta_data(self):
>         return dict(
>             eccentricity=...
>         )
> ```
> 
> We could also put in the information needed to reconstruct the `WaveformGenerator` for that class.
> 
> I'm interested to get thoughts from a wider group on this.

I like this suggestion, and would be in favour of adding `meta_data` to the generators.

lorenzopompili00:
Thanks for taking a look and for the feedback @ColmTalbot @mj-will! I agree your suggestion is cleaner, I have made changes to address it.

PR #1030, opened 2025-12-23
FEAT: add option to crop whitened time series

ColmTalbot:
This PR implements the proposed modification to the likelihood from https://iopscience.iop.org/article/10.1088/1361-6382/ae1ac7/meta.

The main new feature is adding a `crop_duration` to the `InterferometerStrainData` and some accompanying utility code.

I modified how the `GravitationalWaveTransient` and `BasicGravitationalWaveTransient` call the data. **I did not have to change the reference likelihood tests**, so there is good evidence that this doesn't impact the numerical values when not using the cropping.

I suspect there is a performance penalty due to repeatedly whitening the data. There should be a way to cache this calculation, it just requires some cleverness about when the time/frequency masks might have changed.

- I'm not sure at this time how this should interact with accelerated likelihood, so I've left them alone.
- ~I also haven't validated how this interacts with time/calibration marginalization.~


test/gw/detector/interferometer_test.py
```diff
@@ -646,9 +628,17 @@ def _check_time_series_whiteness(self, time_series):
...
 
+    @property
+    def frequency_mask(self):
+        frequencies = bilby.core.utils.series.create_frequency_series(
+            duration=self.ifo.cropped_duration, sampling_frequency=self.sampling_frequency
+        )
+        return (self.ifo.minimum_frequency <= frequencies) & (
+            frequencies <= self.ifo.maximum_frequency
+        )
+
```
mj-will: We should probably have a utility function that does this.

ColmTalbot: Done

bilby/gw/detector/interferometer.py
```diff
@@ -603,11 +605,9 @@ def inner_product(self, signal):
         =======
         float: The optimal signal to noise ratio possible squared
```
GregoryAshton: While we are here, can we update the docstring?

bilby/gw/detector/interferometer.py
```diff
@@ -586,10 +591,7 @@ def optimal_snr_squared(self, signal):
...
         """
-        return gwutils.optimal_snr_squared(
-            signal=signal[self.strain_data.frequency_mask],
-            power_spectral_density=self.power_spectral_density_array[self.strain_data.frequency_mask],
-            duration=self.strain_data.duration)
+        return (abs(self.whiten_frequency_series(signal))**2).sum()
```
GregoryAshton, line 609: Is there not a factor of 4 missing here? It may be helpful to future readers if a reference is given if one exists

bilby/gw/likelihood/base.py
```diff
@@ -278,63 +278,55 @@ def calculate_snrs(self, waveform_polarizations, interferometer, return_array=Tr
...
         if 'recalib_index' in parameters:
-            signal[_mask] *= self.calibration_draws[interferometer.name][int(parameters['recalib_index'])]
+            signal *= self.calibration_draws[interferometer.name][int(parameters['recalib_index'])]
 
-        d_inner_h = interferometer.inner_product(signal=signal)
-        optimal_snr_squared = interferometer.optimal_snr_squared(signal=signal)
+        whitened_signal = interferometer.whiten_frequency_series(signal)
+
+        d_inner_h = (interferometer.whitened_frequency_domain_strain.conjugate() * whitened_signal).sum()
+        optimal_snr_squared = (abs(whitened_signal)**2).sum()
```
GregoryAshton, line 295: Can this not use the implemented method?

bilby/gw/likelihood/base.py
```diff
@@ -391,12 +383,9 @@ def priors(self, priors):
     def _calculate_noise_log_likelihood(self):
         log_l = 0
         for interferometer in self.interferometers:
-            mask = interferometer.frequency_mask
-            log_l -= noise_weighted_inner_product(
-                interferometer.frequency_domain_strain[mask],
-                interferometer.frequency_domain_strain[mask],
-                interferometer.power_spectral_density_array[mask],
-                self.waveform_generator.duration) / 2
+            log_l -= (
```
GregoryAshton, line 397: No need to respond to this, but I will say I preferred how the log-likelihood was previously directly readable from standard definitions. The new implementation feels like a different way of thinking about it.

GregoryAshton (commented):
Overall looks good and I'm supportive, just a few things I wanted to check.

bilby/gw/detector/interferometer.py
ColmTalbot: The whitening methods include the normalization factor so that these are as straightforward as this.

bilby/gw/likelihood/base.py
ColmTalbot: It could, but if the whitening is expensive (which the new method is a little), it's more efficient to only whiten once.

bilby/gw/likelihood/base.py
ColmTalbot: I'm divided on this. I agree the old way is nice and explicit, but it assumes one particular whitening method, and this PR adds a different method.

bilby/gw/detector/interferometer.py
ColmTalbot:
```suggestion
        float:
            The noise-weighted inner product between the passed signal
            and the data stored in the :code:`Interferometer`.
```

bilby/gw/detector/interferometer.py
```diff
@@ -586,10 +591,7 @@ def optimal_snr_squared(self, signal):
         =======
         float: The optimal signal to noise ratio possible squared
```
ColmTalbot:
```suggestion
        float: The optimal signal-to-noise of the signal squared
```

bilby/gw/detector/interferometer.py
GregoryAshton: Okay

bilby/gw/likelihood/base.py
GregoryAshton: That makes sense, then if the idea is one could have different whitening methods the new implementation is cleaner.

bilby/gw/detector/interferometer.py
```diff
@@ -680,7 +676,61 @@ def whiten_frequency_series(self, frequency_series : np.array) -> np.array:
...
-        return frequency_series / (self.amplitude_spectral_density_array * np.sqrt(self.duration / 4))
+        if self.crop_duration == 0:
+            return np.nan_to_num(frequency_series / (
+                self.amplitude_spectral_density_array
+                * np.sqrt(self.duration / 4)
+            )) * self.frequency_mask
+        else:
+            return self.whiten_and_crop(frequency_series=frequency_series)
+
+    def whiten_and_crop(self, frequency_series : np.array) -> np.array:
```
asb5468: It seems like this procedure is more fundamental than suggested by its being a method of the interferometer class. The nice thing about using the gwutils SNR functions was that you could calculate SNRs using them without necessarily initializing an interferometer object. Is it possible to pull this out to the utils and just call it here?

ColmTalbot: Done

bilby/gw/utils.py
```diff
@@ -228,6 +228,104 @@ def overlap(signal_a, signal_b, power_spectral_density=None, delta_frequency=Non
...
+    Returns
+    =======
+    np.ndarray
+        The whitened frequency series
+    """
+    whitened = frequency_series / amplitude_spectral_density
+    return np.nan_to_num(whitened) * frequency_mask * (4 / duration)**0.5
+
+
+def whiten_and_crop(frequency_series, amplitude_spectral_density, frequency_mask, time_mask, duration):
```
GregoryAshton: Should the name state what input it accepts? E.g. like 'frequency_domain_whiten' to make it clear it is acting on the frequency-domain input?

bilby/gw/detector/interferometer.py
```diff
@@ -718,7 +727,9 @@ def get_whitened_time_series_from_whitened_frequency_series(
 
         whitened_time_series = (
             np.fft.irfft(whitened_frequency_series)
-            * np.sqrt(np.sum(self.frequency_mask)) / frequency_window_factor
+            * np.sqrt(np.sum(self.frequency_mask))
+            / frequency_window_factor
+            * self.time_mask.mean()**0.5
```
GregoryAshton: Why does this use both `np.sqrt` and `**0.5`. It would probably be easier to read if it was consistent.

bilby/gw/detector/interferometer.py
```diff
@@ -680,7 +675,21 @@ def whiten_frequency_series(self, frequency_series : np.array) -> np.array:
         frequency_series : np.array
             The frequency series, whitened by the ASD
         """
-        return frequency_series / (self.amplitude_spectral_density_array * np.sqrt(self.duration / 4))
+        if self.crop_duration == 0:
```
GregoryAshton: What happens (or can) crop_duration be "None" or `None` etc? This would silently fail and then trigger the else clause, likely also erroring, but could be confusing?

bilby/gw/likelihood/base.py
```diff
@@ -278,63 +278,55 @@ def calculate_snrs(self, waveform_polarizations, interferometer, return_array=Tr
             interferometer=interferometer,
             parameters=parameters,
         )
-        _mask = interferometer.frequency_mask
 
         if 'recalib_index' in parameters:
```
GregoryAshton, line 288: Why does the signal no longer require masking here?

GregoryAshton (commented):
A few more comments after another look.

bilby/gw/detector/interferometer.py
ColmTalbot: This would fail in the `crop_duration.setter` in the `InterferometerStrainData` with a reasonably explicit error message.

bilby/gw/detector/interferometer.py
ColmTalbot: I think I was trying to avoid explicitly using numpy, but that's unavoidable since we need an fft and sqrt is sometimes meaningfully faster.

```suggestion
            * np.sqrt(self.time_mask.mean())
```

bilby/gw/likelihood/base.py
ColmTalbot: Unclear, is this tested?

bilby/gw/utils.py
ColmTalbot: I can make this change, but it's pretty clear from the signature and docstring.

bilby/gw/detector/interferometer.py
```diff
@@ -614,13 +615,13 @@ def inner_product(self, signal):
...
+            and the data stored in the :code:`Interferometer`.
         """
-        return gwutils.noise_weighted_inner_product(
-            aa=signal[self.strain_data.frequency_mask],
-            bb=self.strain_data.frequency_domain_strain[self.strain_data.frequency_mask],
-            power_spectral_density=self.power_spectral_density_array[self.strain_data.frequency_mask],
-            duration=self.strain_data.duration)
+        whitened_signal = self.whiten_frequency_series(signal)
+        whitened_data = self.whitened_frequency_domain_strain
+        return (whitened_signal.T * whitened_data.conj()).sum()
```
asb5468, line 624: Why does the whitened signal need to be transposed? In the function below, `template_template_inner_product`, it is not.

bilby/gw/detector/interferometer.py
asb5468, line 732: Where does this extra factor of `self.time_mask.mean()**0.5` come from? Can you point to a specific equation in https://arxiv.org/pdf/2508.11091?

bilby/gw/detector/strain_data.py
```diff
@@ -33,6 +35,11 @@ def __init__(self, minimum_frequency=0, maximum_frequency=np.inf,
             This corresponds to alpha * duration / 2 for scipy tukey window.
         notch_list: bilby.gw.detector.strain_data.NotchList
             A list of notches
+        crop_duration: float | tuple
+            The duration of data to crop at the beginning/end of the segment
+            to avoid whitening artifacts. If a float, that duration is excluded
```
asb5468, line 40: The docstring here should add the units for this argument. I'm assuming it's seconds?

bilby/gw/detector/strain_data.py
```diff
@@ -135,6 +145,33 @@ def notch_list(self, notch_list):
             raise ValueError("notch_list {} not understood".format(notch_list))
         self._frequency_mask_updated = False
 
+    @property
+    def crop_duration(self):
+        """
+        The duration of data to crop at the beginning/end of the segment
```
asb5468, line 151: Add the units of this property to the docstring. Also, `crop_duration` and `cropped_duration` are really similar and I'm leery of them getting mixed up. Perhaps `crop_duration` could be renamed to `crop_time`?

bilby/gw/detector/strain_data.py
```diff
@@ -145,20 +182,69 @@ def frequency_mask(self):
...
+
+        cropped_frequencies = create_frequency_series(
+            duration=self.cropped_duration,
+            sampling_frequency=self.sampling_frequency
+        )
+        self._cropped_frequency_mask = calculate_frequency_mask(
+            cropped_frequencies
+        )
+
+        self._frequency_mask_updated = True
```
asb5468, line 208: It seems like `self._frequency_mask_updated` now controls whether both `frequency_mask` and `cropped_frequency_mask` are updated. Is there ever a situation where a user would want to update one but not the other?

bilby/gw/likelihood/base.py
```diff
@@ -277,63 +277,55 @@ def calculate_snrs(self, waveform_polarizations, interferometer, *, return_array
...
-            signal[_mask] *= self.calibration_draws[interferometer.name][int(parameters['recalib_index'])]
+            signal *= self.calibration_draws[interferometer.name][int(parameters['recalib_index'])]
```
asb5468, line 282: Why is this no longer being applied only to the masked frequencies?

bilby/gw/likelihood/base.py
```diff
@@ -277,63 +277,55 @@ def calculate_snrs(self, waveform_polarizations, interferometer, *, return_array
...
+                (self.number_of_response_curves, 1)
+            ).T
 
-            d_inner_h_integrand[_mask] *= self.calibration_draws[interferometer.name].T
+            d_inner_h_integrand[interferometer.frequency_mask] *= self.calibration_draws[interferometer.name].T
 
-            d_inner_h_array = 4 / self.waveform_generator.duration * np.fft.fft(
-                d_inner_h_integrand[0:-1], axis=0
-            ).T
+            d_inner_h_array = np.fft.fft(d_inner_h_integrand[:-1], axis=0).T
```
asb5468, line 305: Why is the 4/T normalization no longer applied here?

asb5468: Ah, is it because it's already applied within `interferometer.whitened_frequency_domain_strain` and `whitened_signal`?

test/gw/detector/interferometer_test.py
```diff
@@ -302,27 +303,6 @@ def test_inject_signal_raises_value_error(self):
         with self.assertRaises(ValueError):
             self.ifo.inject_signal(injection_polarizations=None, parameters=None)
 
-    def test_optimal_snr_squared(self):
```
asb5468, line 305: Why is this test removed?

asb5468 (commented):
A few more general questions/comments:
1. In https://arxiv.org/pdf/2508.11091, there are two options presented for how to treat the template, "Signal models can then either be generated in the frequency domain with frequency spacing 1/T and whitened before subtracting from the whitened and cropped frequency-domain data or generated with the full duration T + 2∆ and processed using the same procedure as the data." Do I understand correctly that this MR has implemented the first option?
2. The code itself seems fine, but it's hard to tell definitively whether there are any unintended consequences, probably because I'm not familiar with all the review tests, and right now they are failing. Have you done any more extensive test runs that you could share, like a GW150914 analysis with the new likelihood with no cropping to make sure the results are unchanged?

bilby/gw/detector/interferometer.py
ColmTalbot: I suspect this is a typo, it likely wouldn't do anything for one dimensional arrays which I think we have an implicit expectation this is, but I'll remove it.

bilby/gw/detector/interferometer.py
ColmTalbot: Note for myself later, it is (10), this is $\sqrt{\beta}$.

bilby/gw/detector/strain_data.py
ColmTalbot: I don't think so, I feel like we should require that they are kept synchronized.

bilby/gw/likelihood/base.py
ColmTalbot: Yup

test/gw/detector/interferometer_test.py
ColmTalbot: I removed this because it doesn't directly call the `bilby.gw.utils.noise_weighted_inner_product` so the test is no longer at all relevant.
There are other tests that directly use this method, so I think we still have sufficient coverage.

bilby/gw/detector/interferometer.py
ColmTalbot: And this was a typo to start with.

bilby/gw/detector/strain_data.py
ColmTalbot: Sure

bilby/gw/likelihood/base.py
ColmTalbot: Related to Greg's question above. I think was a change that is required for array backend consistency where after that `Interferometer.get_detector_response` returns the full array and not the masked version.

PR #1034, opened 2026-01-19
FEAT: Subclass of GWSignalWaveformGenerator for eccentric waveforms

adivijaykumar:
Adds a subclass of GWSignalWaveformGenerator for eccentric waveforms.

I realize that avoiding subclassing/writing source models was one of the reasons we moved away from the old way of calling waveforms. I see these subclasses as "convenience functions", and since it doesn't duplicate a lot of code I am in favour of adding them. But I can also see arguments to the contrary.

### DIFF
--- a/bilby/gw/waveform_generator.py
+++ b/bilby/gw/waveform_generator.py
@@ -581,8 +581,9 @@ class GWSignalWaveformGenerator(WaveformGenerator):
 
 class EccentricGWSignalWaveformGenerator(GWSignalWaveformGenerator):
     """
-    Subclass that initializes an eccentric GW Signal Waveform Generator.
-    See documentation of `GWSignalWaveformGenerator` for more information
+    Subclass that initializes an eccentric GWSignal Waveform Generator.
+    Equivalent to `GWSignalWaveformGenerator` with `eccentric=True`.
+    See documentation of `GWSignalWaveformGenerator` for more information.
     """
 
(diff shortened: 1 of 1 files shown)


bilby/gw/waveform_generator.py
```diff
@@ -579,6 +579,17 @@ def _parameters_from_source_model(self):
         return self._all_parameters.difference(self.defaults.keys())
 
 
+class EccentricGWSignalWaveformGenerator(GWSignalWaveformGenerator):
+    """
+    Subclass that initializes an eccentric GWSignal Waveform Generator.
+    Equivalent to `GWSignalWaveformGenerator` with `eccentric=True`.
+    See documentation of `GWSignalWaveformGenerator` for more information.
```
ColmTalbot, line 586: I think it's worth specifying that this doesn't do any kind of checking that the requested model actually supports eccentricity.

ColmTalbot (commented):
I'm happy to add this as there seems to be a desire for it, although I don't really see how it is preferable to passing an argument.

mj-will:
We discussed this and my understanding is we agreed we wouldn't include this since the number of classses will quickly balloon when one wants to include precession, tides etc. @adivijaykumar does that sound right to you?

mj-will:
Closing this as discussed on the bilby dev call on 30/04/2026

PR #1039, opened 2026-02-09
Sample within constraints compatible with numpy>=2

noahewolfe:
This is an extremely minor change. I developed this starting with version `2.7.1` of bilby. Tested with `numpy` 2.4.1.

Prior to `numpy` 1.25.0, size 1 arrays could be treated as scalars; this was then deprecated (https://numpy.org/doc/2.1/release/1.25.0-notes.html).

Sometimes, calling `PriorDict.evaluate_constraints` returns a size-1 array. This can happen when it is called through `PriorDict.sample_subset_constrained` called with `size=1`. In that case, `n_valid_samples += int(is_valid)` (line 461 of `core/prior/dict.py`) fails as `is_valid` has a size of 1 and cannot be cast to int.

I have seen this occur when using the `dynesty` sampler wrapped in `bilby`, when drawing initial points from the prior.

We fix this by appropriately casting the output of `evaluate_constraints`; if it is size 1, it gets squeezed to a scalar which can then be cast to `int`. Otherwise, the shape is not modified.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,7 +1,10 @@
 name: Publish Python distribution to PyPI
+
 on:
-  release:
-    types: [published]
+  - push
+    branches:
+      - main
+  - release
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


noahewolfe:
> Hi @noahewolfe it looks like this has picked up various changes, can you rebase your branch so we can inspect the changes more cleanly?

... yeah, I think I did that backwards. I'm going to close this for now and make a clean pull request.

PR #1040, opened 2026-02-11
MAINT: Advance deprecations before bilby v3

ColmTalbot:
This PR makes a few changes to the deprecated behaviour slated for removal in bilby v3 (parameters as state and several samplers) for inclusion in v2.8.

Most of the changes are making sure that our test suite exclusively does not use parameters as state. I also set an environment variable in our unit tests to ensure this is enforced.

There was some mocking behaviour that didn't like missing the signature, which is used to determine if likelihood are compliant with passing parameters not as state so I had to remove that from the tests.

### DIFF
--- a/bilby/gw/likelihood/roq.py
+++ b/bilby/gw/likelihood/roq.py
@@ -1186,14 +1186,15 @@ class ROQGravitationalWaveTransient(GravitationalWaveTransient):
         for interferometer in self.interferometers:
             snrs += self.calculate_snrs(
                 waveform_polarizations=signal_polarizations,
-                interferometer=interferometer
+                interferometer=interferometer,
+                parameters=parameters,
             )
         d_inner_h = snrs.d_inner_h_array
         h_inner_h = snrs.optimal_snr_squared
 
         if self.distance_marginalization:
             time_log_like = self.distance_marginalized_likelihood(
-                d_inner_h, h_inner_h)
+                d_inner_h, h_inner_h, parameters=parameters)
         elif self.phase_marginalization:
             time_log_like = ln_i0(abs(d_inner_h)) - h_inner_h.real / 2
         else:


mj-will:
@ColmTalbot should we add deprecation warnings to all the samplers we plan to remove? I guess we leave dynesty and possibly emcee?

bilby/core/sampler/ptemcee.py
```diff
@@ -343,7 +351,8 @@ def get_pos0_from_minimize(self, minimize_list=None):
         def neg_log_like(params):
             """Internal function to minimize"""
             try:
-                parameters = {key: val for key, val in zip(minimize_list, params)}
+                parameters = copy.copy(draw)
+                parameters.update({key: val for key, val in zip(minimize_list, params)})
```
mj-will, line 350: I might be missing something, but where is `draw` defined?

test/gw/likelihood/marginalization_test.py
```diff
@@ -541,36 +542,3 @@ def test_calibration_marginalization(self, time):
             non_marg_ln_ls.append(non_marginalized.log_likelihood_ratio(parameters))
         non_marg_ln_l = logsumexp(non_marg_ln_ls, b=1 / 100)
         self.assertAlmostEqual(marg_ln_l, non_marg_ln_l)
-
-    @parameterized.expand(["no_time", "time"])
-    def test_calibration_marginalization_state(self, time):
```
mj-will, line 546: Why is this being removed?

bilby/core/sampler/ptemcee.py
```diff
@@ -343,7 +351,8 @@ def get_pos0_from_minimize(self, minimize_list=None):
         def neg_log_like(params):
             """Internal function to minimize"""
             try:
-                parameters = {key: val for key, val in zip(minimize_list, params)}
+                parameters = copy.copy(draw)
```
mj-will, line 349: I'm probably missing something, but where is `draw` defined?

test/gw/likelihood/marginalization_test.py
ColmTalbot: This is specifically testing the calibration marginalized likelihood using parameters as state.
It took me a while to work that out from the name...

bilby/core/sampler/ptemcee.py
ColmTalbot: [Here](https://github.com/bilby-dev/bilby/pull/1040/changes#diff-d2101e22c878eb316de1a502418511ba9ba5e39cebe5961783b9e7c765fd5924R370), clear as mud.

bilby/core/sampler/ptemcee.py
ColmTalbot: I'm going to resolve this as I think it's a duplicate.

bilby/core/sampler/ptemcee.py
mj-will: Indeed, clear as mud.

mj-will (approved):

mj-will (commented):
Removing my approval since I realised the CI is failing, but not due the LALSuite issues.

PR #1041, opened 2026-02-12
MAINT: move LikePriorEvaluator to base sampler

mj-will:
Move the `LikePriorEvaluator` class to base sampler.

This allows external sampler plugins to use the class without depending on another sampler plugin. Ideally, I'd get this in for 2.8.0 since it would simplify moving the samplers.

I'm open for suggestions if anyone has thoughts on where to put this.


ColmTalbot (approved):
Sounds good to me

PR #1045, opened 2026-02-17
BUG: Fix sampler_kwargs mutation in dynesty3_utils

ColmTalbot:
I noticed that there was bad efficiency running the ACT walk in parallel because the chains were getting out of sync which it turns out is because we were mutating a dictionary.

### DIFF
--- a/bilby/core/sampler/dynesty3_utils.py
+++ b/bilby/core/sampler/dynesty3_utils.py
@@ -1,4 +1,3 @@
-import os
 import warnings
 from collections import namedtuple
 
@@ -337,12 +336,9 @@ class ACTTrackingEnsembleWalk(BaseEnsembleSampler):
     def sample(args):
         cache = ACTTrackingEnsembleWalk._cache
         if args.kwargs.get("rebuild", False):
-            logger.debug(
-                f"Force rebuilding cache with {len(cache)} remaining on {os.getpid()}"
-            )
+            logger.debug(f"Force rebuilding cache with {len(cache)}.")
             cache.clear()
         if len(cache) == 0:
-            logger.debug(f"Rebuilding cache on {os.getpid()}")
             ACTTrackingEnsembleWalk.build_cache(args)
 
         while len(cache) > 0 and cache[0][2] < args.loglstar:


bilby/core/sampler/dynesty3_utils.py
```diff
@@ -337,7 +337,9 @@ def sample(args):
         cache = ACTTrackingEnsembleWalk._cache
         if args.kwargs.get("rebuild", False):
             logger.debug(f"Force rebuilding cache with {len(cache)} remaining")
+            ACTTrackingEnsembleWalk._cache = list()
             ACTTrackingEnsembleWalk.build_cache(args)
```
mj-will: Is it worth adding some logic and/or an additional argument to `build_cache` that forces it reset the cache. The current behaviour seems quite opaque.

ColmTalbot: I didn't add it to the method since I think there are difficulties with dealing with the static method. I think the new version tells a cleaner story though. Let me know if you have a different suggestion.

I also added some more logging details, I'm not convinced anyone but me ever looks at these, but I think they're useful.

bilby/core/sampler/dynesty3_utils.py
```diff
@@ -336,17 +337,24 @@ def tune(self, tuning_info, update=True):
...
-            logger.debug(f"Force rebuilding cache with {len(cache)} remaining")
-            ACTTrackingEnsembleWalk.build_cache(args)
-        elif len(cache) == 0:
+            logger.debug(
+                f"Force rebuilding cache with {len(cache)} remaining on {os.getpid()}"
+            )
+            cache.clear()
+        if len(cache) == 0:
+            logger.debug(f"Rebuilding cache on {os.getpid()}")
             ACTTrackingEnsembleWalk.build_cache(args)
```
mj-will, line 342: This looks better to me. I see cache is not assigned here, is that because the 'pointer' on line 338 still points to the correct object in memory?

bilby/core/sampler/dynesty3_utils.py
```diff
@@ -336,17 +337,24 @@ def tune(self, tuning_info, update=True):
     def sample(args):
         cache = ACTTrackingEnsembleWalk._cache
         if args.kwargs.get("rebuild", False):
-            logger.debug(f"Force rebuilding cache with {len(cache)} remaining")
-            ACTTrackingEnsembleWalk.build_cache(args)
-        elif len(cache) == 0:
+            logger.debug(
+                f"Force rebuilding cache with {len(cache)} remaining on {os.getpid()}"
```
mj-will: Just noting that technically, using f-strings is slower here since `os.getpid` will always be called even if the logging level is higher than debug. See e.g. https://docs.astral.sh/ruff/rules/logging-f-string/

I'm guilty of this myself and I think it's everywhere in the code, but perhaps something to discuss more generally.

bilby/core/sampler/dynesty3_utils.py
ColmTalbot: Yeah, this relies pretty heavily on a process-global class attribute for the cache.

bilby/core/sampler/dynesty3_utils.py
ColmTalbot: I removed this, it's a pretty minimal cost, but I think it probably isn't worth it.

PR #1046, opened 2026-02-18
MAINT: remove external samplers

### LINKED ISSUE #795: Transitioning to sampler plugins
In GitLab by @git.ligo:michael.williams on Mar 25, 2024, 15:30

I've opened this issue to track changes that should be made as we migrate samplers out of the main code and into their own packages. Hopefully others can edit this as they identify things.

**Samplers**

We haven't agreed a policy for moving the samplers, but I believe nessai will be the first, followed by pypolychord. As we decided on more samplers, we can add them here.

In order of decreasing priority:

- [x] nessai
- [x] pypolychord

**Examples**

Some examples may need updating because of the changes

- [ ] `examples/core_examples/alternative_samplers/linear_regression_pymc_custom_likelihood.py` (should be made if/when moving pymc)

mj-will:
Remove the external samplers for 3.0.

I see a few open questions:

- How do we want to handle the examples?
    - For now I've left the examples that use other samplers but added an explicit error.
- How should the tests work if external samplers are installed?
    - At the moment, they detect all available samplers, including any external samplers that are installed.
- One of the doc notebooks uses external samples, should these be installed for the entire CI or just the docs CI?
- Should we still provide `sampler_requirements.txt`?


containers/environment.yml
```diff
@@ -46,17 +46,7 @@ dependencies:
...
-  - ptemcee
-  - pymultinest
-  - ultranest
-  - cpnest
-  - kombine
-  - zeus-mcmc
   - pytorch
-  - pymc>=5.9
-  - nessai
-  - ptmcmcsampler
```
ColmTalbot, line 57: There are probably other packages that can be removed from here, but I imagine it will be a pain to go through and figure out which.

ColmTalbot (commented):
> - How do we want to handle the examples?
>   - For now I've left the examples that use other samplers but added an explicit error.

This looks good to me, neater than what we had before.

> - How should the tests work if external samplers are installed?
>   - At the moment, they detect all available samplers, including any external samplers that are installed.

Probably, we should only test the ones we package.
It would be great if we can expose this test in some way that plugins can use it directly.

> - One of the doc notebooks uses external samples, should these be installed for the entire CI or just the docs CI?

Yeah, I think we should have at least a few of those available, maybe in the full CI.
It might be worth changing which ones are used, restrict it to maintained, easy to install ones?

> - Should we still provide sampler_requirements.txt?

I'm leaning to no, but could go either way.

mj-will:
I've changed the samplers to: dynesty, emcee, nessai, nestle and zeus since these are all easy to install.

The plugins are installed with pip for now but once they're on conda we can update that.#

I've changed the tests so they only run the native samplers. I'll have a look at how we could expose the tests, but I think that might be something for another PR.

mj-will:
This is going to need a rebuild of the containers but I forget how to trigger that. I may need to more this PR to the main repo. @ColmTalbot can you remind me?

mj-will:
Superseded by https://github.com/bilby-dev/bilby/pull/1090

PR #1049, opened 2026-02-19
DEP: remove support for dynesty v2

ColmTalbot:
Drop support for older versions of dynesty for version 3.


mj-will (approved):
LGTM

bilby/core/sampler/dynesty_utils.py
```diff
@@ -1,118 +1,182 @@
...
-
         # update walks to match target naccept
-        accept_prob = max(0.5, blob["accept"]) / self.kwargs["walks"]
+        accept_prob = max(0.5, tuning_info["accept"]) / self.sampler_kwargs["walks"]
         delay = max(self.nlive // 10 - 1, 0)
-        n_target = getattr(_SamplingContainer, "naccept", 60)
-        self.walks = (self.walks * delay + n_target / accept_prob) / (delay + 1)
-        self.kwargs["walks"] = min(int(np.ceil(self.walks)), _SamplingContainer.maxmcmc)
+        self.scale = tuning_info["accept"]
+        self.walks = (self.walks * delay + self.naccept / accept_prob) / (delay + 1)
```
GregoryAshton, line 125: Is this changing the way the walks is updated? Or am I misunderstanding the diff?

bilby/core/sampler/dynesty_utils.py
```diff
@@ -160,16 +222,25 @@ def __call__(self, args):
             current_v = args.prior_transform(current_u)
             logl = args.loglikelihood(current_v)
 
-        blob = {
+        sampling_info = {
```
GregoryAshton, line 225: Is there a reason the scale is no longer tracked?

bilby/core/sampler/dynesty_utils.py
```diff
@@ -185,32 +256,113 @@ class ACTTrackingRWalk:
...
     _cache = list()
 
-    def __init__(self):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
         self.act = 1
-        self.thin = getattr(_SamplingContainer, "nact", 2)
-        self.maxmcmc = getattr(_SamplingContainer, "maxmcmc", 5000) * 50
+        self.thin = kwargs.get("nact", 2)
+        self.maxmcmc = kwargs.get("maxmcmc", 5000) * 50
```
GregoryAshton, line 263: Should we make this 50 factor configurable?

bilby/core/sampler/dynesty_utils.py
```diff
@@ -292,16 +453,18 @@ def build_cache(self):
...
             samples=np.array(u_list),
             most_failures=most_failures,
         )
         reject += nfail
-        blob = {"accept": accept, "reject": reject, "scale": args.scale}
-        iact = self.integer_act
-        thin = self.thin * iact
+        blob = {"accept": accept, "reject": reject, "act": act}
+        iact = ACTTrackingEnsembleWalk.integer_act(act)
+        thin = args.kwargs.get("thin", 2) * iact
```
GregoryAshton, line 465: Is this setting a default `thin` of 2? Is that new or just  change of where the default is set?

GregoryAshton (commented):
Overall looks good. Review tests would be needed to confirm sampling has not been impacted, but I don't see anything of significant concern.

bilby/core/sampler/dynesty_utils.py
ColmTalbot: I'm pretty sure the logic is the same, it's just some of the variable names that are different. The appearance of changes is mostly because I renamed the existing dynesty3_utils (compare with https://github.com/bilby-dev/bilby/blob/main/bilby/core/sampler/dynesty3_utils.py#L122-L126.)

bilby/core/sampler/dynesty_utils.py
ColmTalbot: This is included in the older version as a legacy thing from when we used to update the scale of proposals during the MCMC chain. I think we decided that that breaks things and so we don't use it anymore.

bilby/core/sampler/dynesty_utils.py
ColmTalbot: We could, we would need to keep it matched to this line, https://github.com/bilby-dev/bilby/blob/main/bilby/core/sampler/dynesty3_utils.py#L377.

I'm going to suggest not in this PR to avoid introducing meaningful changes in the refactoring commit.

bilby/core/sampler/dynesty_utils.py
ColmTalbot: Just a difference in where it is defined between the two versions (https://github.com/bilby-dev/bilby/blob/main/bilby/core/sampler/dynesty_utils.py#L190).

mj-will:
As discussed on the call, there's need to be some coordination so that https://github.com/bilby-dev/bilby/issues/1065 is addressed.

GregoryAshton (approved):
Looks good

PR #1050, opened 2026-02-19
DEP: remove ability to use parameters as state

ColmTalbot:
Likelihoods (and the waveform generator) will no longer allow a `.parameters` attribute to be used.

### DIFF
--- a/bilby/core/likelihood.py
+++ b/bilby/core/likelihood.py
@@ -9,86 +9,10 @@ from scipy.stats import multivariate_normal
 
 from .utils import infer_parameters_from_function, infer_args_from_function_except_n_args, logger
 
-PARAMETERS_AS_STATE = os.environ.get("BILBY_ALLOW_PARAMETERS_AS_STATE", "WARN")
-for msg in [
-    "No parameters provided in likelihood call",
-    "Using parameters as state",
-    "Parameter attribute queried",
-    "Settings non-trivial"
-]:
-    warnings.filterwarnings("once", message=msg)
-
-
-def set_parameters_as_state(level):
-    global PARAMETERS_AS_STATE
-
-    level = str(level).upper()
-    if level not in ["TRUE", "FALSE", "WARN"]:
-        raise ValueError("Allowed levels for parameters as state are 'TRUE', 'FALSE', and 'WARN'")
-    PARAMETERS_AS_STATE = level
-
-
-def _fallback_to_parameters(obj, parameters):
-
-    if parameters is None:
-        msg = (
-            f"No parameters provided in likelihood call, falling back to values stored in {obj}. "
-            "This is deprecated behaviour  and will be removed in Bilby version 3. "
-            "See https://bilby-dev.github.io/bilby/parameters for more details."
-        )
-        if PARAMETERS_AS_STATE == "FALSE":
-            raise LikelihoodParameterError(msg)
-        elif PARAMETERS_AS_STATE == "WARN":
-            warnings.warn(msg, FutureWarning)
-        else:
-            logger.debug(msg)
-        parameters = copy.deepcopy(obj.parameters)
-
-    return parameters
-
-
-def _safe_likelihood_call(likelihood, parameters=None, use_ratio=False):
-    if use_ratio:
-        method = likelihood.log_likelihood_ratio
-    else:
-        method = likelihood.log_likelihood
-
-    if "parameters" in inspect.signature(method).parameters:
-        logl = method(parameters=parameters)
-    else:
-        if PARAMETERS_AS_STATE == "FALSE":
-            raise LikelihoodParameterError(
-                f"Unable to call {likelihood} with {parameters} as an argument"
-            )
-        elif PARAMETERS_AS_STATE == "WARN":
-            warnings.warn(
-                f"Using parameters as state for {likelihood}. "
-                "This is deprecated behaviour  and will be removed in Bilby version 3. "
-                "See https://bilby-dev.github.io/bilby/parameters for more details.",
-                DeprecationWarning,
-            )
-        likelihood.parameters.update(parameters)
-        logl = method()
-    return logl
-
 
 class Likelihood:
 
-    def __init_subclass__(cls, **kwargs):
-        super().__init_subclass__(**kwargs)
-        if (
-            "parameters" not in inspect.signature(cls.log_likelihood).parameters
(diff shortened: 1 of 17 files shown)


bilby/core/likelihood.py
```diff
@@ -1,138 +1,29 @@
...
-            )
-
-    def __init__(self, parameters=None):
+    def __init__(self):
         """Empty likelihood class to be subclassed by other likelihoods
 
         Parameters
         ==========
         parameters: dict
             A dictionary of the parameter names and associated values
```
mj-will: This doc-string should be updated.

bilby/core/likelihood.py
```diff
@@ -1,138 +1,29 @@
...
-                warnings.warn(msg, FutureWarning)
-        else:
-            parameters = dict()
-        self._parameters = parameters
-
     def __repr__(self):
         return self.__class__.__name__
 
-    def log_likelihood(self, parameters=None):
+    def log_likelihood(self, parameters):
```
mj-will, line 20: I think all the doc-strings should be updated. This could either be done in this PR or a separate one.

bilby/gw/waveform_generator.py
```diff
@@ -131,16 +131,14 @@ def frequency_domain_strain(self, parameters=None):
...
         """ Wrapper to source_model.
 
-        Converts self.parameters with self.parameter_conversion before handing it off to the source model.
+        Converts parameters with self.parameter_conversion before handing it off to the source model.
         Automatically refers to the frequency_domain_source model via INFFT if no frequency_domain_source_model is
         given.
 
         Parameters
         ==========
         parameters: dict, optional
```
mj-will: I don't paramteres should be `optional` anymore.

mj-will (commented):
I couple of minor comments mostly about doc-strings.

bilby/gw/likelihood/base.py
```diff
@@ -250,7 +250,7 @@ def _check_set_duration_and_sampling_frequency_of_waveform_generator(self):
                     "waveform_generator.".format(attribute))
             setattr(self.waveform_generator, attribute, ifo_attr)
 
-    def calculate_snrs(self, waveform_polarizations, interferometer, return_array=True, parameters=None):
+    def calculate_snrs(self, waveform_polarizations, interferometer, *, return_array=True, parameters):
```
asb5468, line 253: What does this `*` do here?

asb5468: Ok, I understand that this means that all arguments that appear after the * must be passed as keyword arguments.

bilby/gw/likelihood/base.py
```diff
@@ -438,8 +435,7 @@ def log_likelihood_ratio(self, parameters=None):
 
         return float(log_l.real)
 
-    def compute_log_likelihood_from_snrs(self, total_snrs, parameters=None):
-        parameters = _fallback_to_parameters(self, parameters)
+    def compute_log_likelihood_from_snrs(self, total_snrs, parameters):
```
asb5468, line 438: I would like to better understand the logic for which of these methods get the `*` added to the definition to require the parameters to be a keyword argument. Why is that not implemented here, for example? Or in `_compute_full_waveform`, `_setup_rho`, and `get_sky_frame_parameters`?

bilby/gw/waveform_generator.py
```diff
@@ -73,8 +73,10 @@ def __init__(self, duration=None, sampling_frequency=None, start_time=0, frequen
             self.waveform_arguments = waveform_arguments
         else:
             self.waveform_arguments = dict()
-        if isinstance(parameters, dict):
-            self.parameters = parameters
+        if parameters is not None:
+            logger.warning(
+                "Non null parameters passed to waveform generator. These will be ignored."
+            )
```
asb5468, line 80: I don't really understand this change. Why are we not setting `self.parameters` in the waveform generator anymore? I thought that change was limited to the likelihood.

examples/core_examples/alternative_samplers/linear_regression_pymc_custom_likelihood.py
```diff
@@ -75,7 +75,7 @@ def __init__(self, x, y, sigma, func):
         """
         super(GaussianLikelihoodPyMC, self).__init__(x=x, y=y, func=func, sigma=sigma)
 
-    def log_likelihood(self, sampler=None, parameters=None):
+    def log_likelihood(self, sampler=None, *, parameters):
```
asb5468, line 85: I don't think the `*` should be here, as this syntax was not enforced in any of the GWTransient likelihood classes' calls to `log_likelihood`.

bilby/gw/likelihood/base.py
ColmTalbot: My goal was to not change the order of the current arguments but remove the default value for parameters.
Here, there are no keyword arguments before `parameters` and so the `*` is not needed to enforce this.

For `calculate_snrs`, the `return_array` argument has a default, so to make this a non-keyword only argument requires one of:

- change the order of arguments. If people are using return_array with a positional argument this will fail with a strange error.
- allow parameters to be positional and have some explicit catch for parameters being `None`. This feels like it would require annoying boilerplate.
- make at least `parameters` keyword only.

I also made `return_array` keyword-only also as this feels like a natural case where it should be a keyword argument.

I am not strongly against the first option and am interested to get other opinions on what users might prefer.
Personally, I would rather not do option 2 unless you are strongly against the other options.

Looking through, there are some cases where I made the `parameters` keyword only where it doesn't really need to be (e.g., `time_marginalized_likelihood`).
I guess I got carried away...

bilby/gw/waveform_generator.py
ColmTalbot: I don't really understand how or if these input parameters are used in practice.
If you have a specific use case that this change would make worse I'm happy to hear it, but in my mind this change is not limited to just the likelihood.

examples/core_examples/alternative_samplers/linear_regression_pymc_custom_likelihood.py
asb5468: I now understand that this is needed here because `sampler` has a default value.

asb5468 (approved):

mj-will (approved):
Thanks for making those changes Colm, LGTM now!

mj-will (approved):

Issue #1047, opened 2026-02-19
BUG: Issue with prior


PawanGu:
**Describe the bug**
Sampler does not work due to prior issue.

**To Reproduce**
Just running GW190425.py from example code
**Error message**
Traceback (most recent call last):
  File "/home/project/bilby_testing/GW150914/GW150914.py", line 101, in <module>
    result = bilby.run_sampler(
             ^^^^^^^^^^^^^^^^^^
  File "/home/miniconda3/envs/lalsuite-env/lib/python3.11/site-packages/bilby/core/sampler/__init__.py", line 273, in run_sampler
    sampler = sampler_class(
              ^^^^^^^^^^^^^^
  File "/home/miniconda3/envs/lalsuite-env/lib/python3.11/site-packages/bilby/core/sampler/dynesty.py", line 238, in __init__
    super(Dynesty, self).__init__(
  File "/home/miniconda3/envs/lalsuite-env/lib/python3.11/site-packages/bilby/core/sampler/base_sampler.py", line 267, in __init__
    self._verify_parameters()
  File "/home/miniconda3/envs/lalsuite-env/lib/python3.11/site-packages/bilby/core/sampler/base_sampler.py", line 432, in _verify_parameters
    theta = self.priors.sample_subset_constrained_as_array(
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/miniconda3/envs/lalsuite-env/lib/python3.11/site-packages/bilby/core/prior/dict.py", line 378, in sample_subset_constrained_as_array
    samples_dict = self.sample_subset_constrained(keys=keys, size=size)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/miniconda3/envs/lalsuite-env/lib/python3.11/site-packages/bilby/core/prior/dict.py", line 461, in sample_subset_constrained
    n_valid_samples += int(is_valid)
                       ^^^^^^^^^^^^^
TypeError: only 0-dimensional arrays can be converted to Python scalars

**Environment (please complete the following information):**
 - Bilby version: 2.7.1
 - Installation method: Conda

mj-will:
Hi @PawanGu this is an issue with numpy 2.4.x, downgrading to an older release should fix this,

mj-will:
[...]

PR #1061, opened 2026-03-12
Fixing #1047

GregoryAshton:
## Issue

[GitHub #1047](https://github.com/lscsoft/bilby/issues/1047)

`sample_subset_constrained` raised a `TypeError` when sampling with `size=None` or `size=1`:

```
TypeError: only 0-dimensional arrays can be converted to Python scalars
```

## Root Cause

`evaluate_constraints` returns a numpy array via `np.ones_like`. When sampling a single point, this is a 1-element array rather than a scalar. The subsequent `int(is_valid)` call on line 464 fails because `int()` only accepts 0-dimensional arrays.

## Fix

Wrap `is_valid` with `np.squeeze` before converting to `int`, reducing the 1-element array to a 0-d scalar:

```python
# before
n_valid_samples += int(is_valid)

# after
n_valid_samples += int(np.squeeze(is_valid))
```

### DIFF
--- a/bilby/core/prior/dict.py
+++ b/bilby/core/prior/dict.py
@@ -461,7 +461,7 @@ class PriorDict(dict):
                 sample = self.sample_subset(keys=keys, size=size)
                 is_valid = self.evaluate_constraints(sample)
                 n_tested_samples += 1
-                n_valid_samples += int(is_valid)
+                n_valid_samples += int(np.squeeze(is_valid))
                 check_efficiency(n_tested_samples, n_valid_samples)
                 if is_valid:
                     return sample


GregoryAshton:
My apologies @ColmTalbot I may have missed your MR that fixed it. If that is the case, please do close this

mj-will:
Agreed on the call we would backport this to 2.7.x

asb5468 (approved):

ColmTalbot (approved):

mj-will:
> I think I added a fix for this in #886 that is backend agnostic. That must be where I'd seen it before. Can we backport the change from #886 in 1103 @mj-will and close this?

Sure, I've backported that fix in dcda70db0865ff2af7cd789627015b396e503576, so closing this.

PR #1064, opened 2026-03-31
BUG: median and error bar method

### LINKED ISSUE #1062: BUG: `get_one_dimensional_median_and_error_bar` returns class instead of instance
**Describe the bug**
method .get_one_dimensional_median_and_error_bar returns class, not instance, and also has different fields from what described in the api documentation

**To Reproduce**
For some CBCResult object `result`, and a parameter, for example, chirp_mass,
```
err = result.get_one_dimensional_median_and_error_bar("chirp_mass")
print(type(err), err)
```
will return
<class 'type'> <class 'bilby.core.result.summary'>

To check fields, I tried
```
print(err._fields)
```
and this returns
('median', 'lower', 'upper', 'string')
as written in api. However,
```
print(err.upper)
```
gives _tuplegetter(2, 'Alias for field number 2')

To get the expected value for 'lower' and 'upper', I had to use 'plus' and 'minus' instead. `print(err.plus)` gives a float value.

**Expected behavior**
the method giving instance, not type, and fields working as written in api

**Suggested solution (optional)**
I've used `inspect.getsource` to see what is the problem, and I found that the object summary is defined to have 'lower' and 'upper' to be fields, while the values themselves are assigned to 'plus' and 'minus'. Also, the returned value is the class summary itself, not the instance defined like `summary(a, b)`. I think simply changing field names and return type will fix this, and maybe I could do it, but this is my first time contributing to an open project, so I'm not sure I could do it right.

**Environment (please complete the following information):**
 - Bilby version: 2.8.0
 - Installation method [e.g. conda, pip, source]: installed by uv in wsl, using Quarto for my project. However I could reproduce this in terminal

**Additional context**

This is the result of upper codes I got in my laptop. Sorry if this is already known issue.

Image

ColmTalbot:
Closes #1062

### DIFF
--- a/bilby/core/result.py
+++ b/bilby/core/result.py
@@ -1055,7 +1055,7 @@ class Result(object):
         fmt = "{{0:{0}}}".format(fmt).format
         string_template = r"${{{0}}}_{{-{1}}}^{{+{2}}}$"
         string = string_template.format(fmt(median), fmt(minus), fmt(plus))
-        return summary(median, minus, upper, string)
+        return summary(median, minus, plus, string)
 
     @latex_plot_format
     def plot_single_density(self, key, prior=None, cumulative=False,
--- a/bilby/core/result.py
+++ b/bilby/core/result.py
@@ -2,7 +2,6 @@ import datetime
 import inspect
 import json
 import os
-from collections import namedtuple
 from copy import copy
 from importlib import import_module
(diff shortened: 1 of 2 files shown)


adivijaykumar (approved):
Looks good to me!

asb5468 (approved):

PR #1067, opened 2026-04-01
Fix and test containers

ColmTalbot:
The CI tests are pretty broken again. I'm not sure if this is due to changes I made or not.

I added some tests to the containers between building and pushing them, so hopefully we won't have broken containers being pushed in the future.

Summary of changes:

- the dockerfile gets `SETUPTOOLS_SCM_IGNORE_DUBIOUS_OWNER=1`. This tells setuptools_scm to ignore the fact that it is being run in a repository the user doesn't own.
- the containers are stored to local storage after being built.
- these local containers run the basic tests (imports and entry points).
- if these tests pass, the last stage uploads the containers to the registry.

I don't understand why the user doesn't own the repository anymore, but this fixes the problem at least.

As usual, if this isn't merged by midnight the old container will be rebuilt and things will start breaking again.

I also reactivated the tests that our command line entry points work.

### DIFF
--- a/.github/workflows/basic-install.yml
+++ b/.github/workflows/basic-install.yml
@@ -37,23 +37,9 @@ jobs:
       - name: List environment
         run: python -m pip list installed
       - name: Test imports
+        run: bash test/ci_test_imports.sh
+      - name: Test entry points
         run: |
-          python -c "import bilby"
-          python -c "import bilby.bilby_mcmc"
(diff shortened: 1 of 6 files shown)


GregoryAshton (approved):
I read over the MR and it looks good to me.

adivijaykumar (approved):

PR #1077, opened 2026-04-17
ENH: introduce combined Prior object

Hen42rik:
I suggest to add a Combined-prior that evaluates the superposition of other bilby prior classes.


fgittins:
This seems like a sensible addition. But I wonder if it would make sense to add this functionality to something like `PriorDict`.

Hen42rik:
@fgittins I have added the ability to initialise from a PriorDict, but I do not quite see in what way the PriorDict class itself would benefit. The use case I have in mind would be some simple superposition of e.g. two gaussians or some other prior on top of a constant background.

On a different note, in order to pass the tests, I have made some more fundamental changes to how the prior class initialises from its representation. I am open to discussions for better ways to handle this. 

Finally, while all tests pass locally, I would greatly appreciate some help to understand why the test_repr-part still fails for Interped priors in the workflow...

Hen42rik:
@fgittins would you be available for review? I am not sure how to proceed with this:)

fgittins:
@Hen42rik, sorry for the radio silence on this. Your change has recently been discussed among the other developers and the consensus is that this would be a helpful feature.

However, there are a couple of changes needed before we can merge this.

1. The `Combined` class name is too abstract and uninformative. This seems to be a mixture model. If it is, then the name should reflect that.
2. There seems to be a bit of code repetition between this class and the existing interpolated prior. Please have a look at refactoring your class to inherit from this.

Once you have made these changes, we'll give it another review and go from there!

PR #1085, opened 2026-05-01
ENH: GWSignal disabled parameter warnings

mj-will:
This PR adds warnings to the GWSignal generator that trigger when parameters that are disabled (e.g. when eccentric=False) are specified when calling the generator.

For now, these are just logger warnings but we could change these to an error if we wanted.

Related to #1034

### DIFF
--- a/bilby/gw/waveform_generator.py
+++ b/bilby/gw/waveform_generator.py
@@ -398,6 +398,30 @@ class GWSignalWaveformGenerator(WaveformGenerator):
         output = {key: 0.0 for key in keys}
         return output
 
+    @property
+    def _disabled_parameters(self):
+        disabled = set()
+        if not self.spinning:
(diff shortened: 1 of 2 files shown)


bilby/gw/waveform_generator.py
```diff
@@ -398,6 +398,30 @@ def defaults(self):
         output = {key: 0.0 for key in keys}
         return output
 
+    @property
+    def _disabled_parameters(self):
```
ColmTalbot, line 402: This is copying quite a bit from the defaults. Can you just use the keys of `self.defaults`?

mj-will: Good point, I'll change that.

ColmTalbot (approved):
Looks good to me now

adivijaykumar (approved):

PR #1090, opened 2026-05-07
DEP: remove external samplers

mj-will:
This PR supersedes https://github.com/bilby-dev/bilby/pull/1046.

It is made from the main repo so the containers can be rebuilt.

See the original PR for previous discussions.


mj-will:
I've triggered a workflow to rebuild the containers: https://github.com/bilby-dev/bilby/actions/runs/25503395807

ColmTalbot (commented):
Looks like some change are still needed in https://github.com/bilby-dev/bilby/blob/remove-external-samplers/test/core/sampler/base_sampler_test.py, you might also pin `arviz<1` until we support the new version.

mj-will:
@ColmTalbot any idea where this dynesty failure is coming from?

mj-will:
I agree the dynesty test probably shouldn't be a required test (I think you might have changed that already?).

adivijaykumar (approved):
Looks good to me, modulo the CI failure.

ColmTalbot (approved):

PR #1091, opened 2026-05-07
MAINT: Initial attempt to support the new arviz API.

ColmTalbot:
See https://python.arviz.org/en/stable/user_guide/migration_guide.html for the migration guide.
I suspect we'll want to support the old version for a while, and if this works, it isn't a large burden.


mj-will (approved):
LGTM, I agree it's worth support both for the time being.

mattpitkin (approved):

PR #1092, opened 2026-05-14
BUG: specify sample_dims for arviz>1

mj-will:
Fix the latest CI failure related to arviz>1

### DIFF
--- a/bilby/core/result.py
+++ b/bilby/core/result.py
@@ -1842,7 +1842,7 @@ class Result(object):
         if az_version < packaging.version.parse("1"):
             azdata = az.from_dict(**az_data_dict)
         else:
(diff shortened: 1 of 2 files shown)


adivijaykumar (approved):

ColmTalbot (approved):
It looks like the core result tests that were failing before have passed so I'm approving.

mj-will:
Ah, the fix in the tests doesn't work for arviz<1. Will fix that now

mj-will:
I think we'll have to forceable merge this and https://github.com/bilby-dev/bilby/pull/1090 since the queue won't work so long as the PyMC test is failing. What do you think @ColmTalbot ?

PR #1096, opened 2026-05-20
ENH: add checks for prior constraints

mj-will:
Some tweaks to prior contraints

- Raise an error if a constraint is present but the parameter key is missing. Currently, it's possible to have constraints that are silently ignored, e.g. because the conversion function doesn't generate them.
- Remove an unreachable if statement. Prior to https://github.com/bilby-dev/bilby/pull/1028, `evaluate_constraints`  returned 1.0 if there were no valid constraints and there was a check for this where the normalization was set to 1. I've removed this for now. Alternatively we could return 1.0 if all values are true but the while should do this anyway.

### DIFF
--- a/bilby/core/prior/dict.py
+++ b/bilby/core/prior/dict.py
@@ -61,19 +61,33 @@ class PriorDict(dict):
         return hash(str(self))
 
     @xp_wrap
-    def evaluate_constraints(self, sample, *, xp=None):
+    def evaluate_constraints(self, sample, *, strict=True, xp=None):
         """Evaluate the constraints for a given sample.
 
         Applies the conversion function to the sample and evaluates the
         constraints on the converted sample.
 
+        Parameters
+        ==========
+        sample: dict
+            Dictionary of parameters used to evaluate the constraints.
+        strict: bool, optional
+            When True, raise if a constraint cannot be evaluated from the
+            provided sample. When False, skip constraints that cannot be
+            derived from a partial sample.
+
         Raises
         ======
         ValueError:
             If a constraint parameter is not present in the sample after
-            conversion.
+            conversion and ``strict`` is True.
         """
-        out_sample = self.conversion_function(sample)
+        try:
+            out_sample = self.conversion_function(sample)
+        except KeyError:
+            if strict:
+                raise
+            out_sample = sample.copy()
         try:
             prob = xp.ones_like(next(iter(out_sample.values())), dtype=bool)
         except TypeError:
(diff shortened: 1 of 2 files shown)


mj-will:
I think we may need to have an option where this doesn't raise an error since it is called in `sample_subset_constrained` and there are cases there where it shouldn't fail. E.g. if evaluating the usual GW prior and the masses aren't present.

mj-will:
> > I think we may need to have an option where this doesn't raise an error since it is called in sample_subset_constrained and there are cases there where it shouldn't fail. E.g. if evaluating the usual GW prior and the masses aren't present.
> 
> We could add this to the CBCPriorDict?

I've added  `strict` option, let me know what you think.

bilby/core/prior/dict.py
```diff
@@ -530,9 +570,6 @@ def normalize_constraint_factor(
     def _estimate_normalization(self, keys, min_accept, sampling_chunk):
         samples = self.sample_subset(keys=keys, size=sampling_chunk)
         keep = np.atleast_1d(self.evaluate_constraints(samples))
-        if len(keep) == 1:
```
ColmTalbot, line 533: Why has this been removed?

ColmTalbot (commented):
It looks like this is a hard condition on the sample, prob, and ln_prob methods.
I don't know that the latter two are correctly handled.
For example, the following is currently valid, but I suspect will fail under this change.
Is that right?

```python
import bilby

priors = bilby.gw.prior.BBHPriorDict
priors.ln_prob(dict(a_1=0.3, tilt_1=0.7))
```

bilby/core/prior/dict.py
mj-will: My understanding was that this was a hacky check for if there aren't any constraints.  Prior to https://github.com/bilby-dev/bilby/pull/1028, `evaluate_constraints` would return `1.0` if there weren't any constraints, so this would skip the rest of this function.

After the changes 1028, that only happens if the `sampling_chunk=1` is, so it's not clear to me this actually serves it's intended purpose.

Looking at this again, I think the lines above this may be redundant as well and this whole thing could be replaced with a `has_constraints` (or similar) property.

mj-will:
> It looks like this is a hard condition on the sample, prob, and ln_prob methods. I don't know that the latter two are correctly handled. For example, the following is currently valid, but I suspect will fail under this change. Is that right?
> 
> ```python
> import bilby
> 
> priors = bilby.gw.prior.BBHPriorDict
> priors.ln_prob(dict(a_1=0.3, tilt_1=0.7))
> ```

Yes, this fails but I think it would fail before as well since the conversion function would fail with a `KeyError` due to the masses missing.

You raise a good point though, I had assumed `ln_prob` should only called with a complete sample but I don't know if that's actually the case.

mj-will:
@ColmTalbot I've had another look at this and changed so that the `strict` option is only enabled by default if all the parameters are present in the sampler. This means partial samples should still work unless `strict` is set manually.

PR #1097, opened 2026-05-29
ENH: support saving/loading numpy.random.SeedSequence

mj-will:
Add support for saving/loading `numpy.random.SeedSequence` objects from results.

Such objects can be present in `bilby.core.utils.random.seed` is called with `SeedSequence`, since this adds the seed to the global meta data.

This can lead to issues saving results now the global meta data is always included.


mj-will:
Note to self: we should add a test where the global meta data has a SeedSequence

ColmTalbot (approved):

GregoryAshton (commented):
LGTM

adivijaykumar (approved):

Issue #1098, opened 2026-06-11
BUG: PriorDict: Missing conversion function will silently ignore Constraints


AlexandreGoettel:
**Describe the bug**
When sampling from a PriorDict that contains Constraints, if that PriorDict does not have a conversion function, the Constraints will silently be ignored. The problem is not the ignoring, but the fact that it is silent.

**To Reproduce**
In the context of a GW231123 analysis with NRSur7dq4, you might have:

```python
priors = bilby.core.prior.PriorDict({"total_mass": bilby.core.prior.Constraint(name='total_mass', minimum=60, maximum=400), "chirp_mass": bilby.gw.prior.UniformInComponentsChirpMass(minimum=85, maximum=194.62, name='chirp_mass'), "mass_ratio": bilby.gw.prior.UniformInComponentsMassRatio(minimum=0.166666667, maximum=1.0, name='mass_ratio')})
priors.sample()  # ---> will contain invalid samples
```

**Error message**
N/A

**Expected behavior** / **Suggested solution (optional)**
I would suggest adding a simple warning in case a constraint is being ignored.

**Environment (please complete the following information):**
 - Bilby version: 2.6.0
 - Installation method [e.g. conda, pip, source]: conda

PR #1099, opened 2026-06-11
Add warning to PriorDict.evaluate_constraints.

AlexandreGoettel:
Addresses #1098

Technically, if the conversion function does not affect some of the constrained parameters they will be ignored too, but to check this would (afaik) require quite some refactoring and I'm not sure it is worth implementing. 

However, having talked to people that use Bilby, all of them have been surprised and worried by this behaviour, so this warning, which will affect all cases in which a prior is defined without an explicit conversion function, is a good step imo.

### DIFF
--- a/bilby/core/prior/dict.py
+++ b/bilby/core/prior/dict.py
@@ -55,6 +55,8 @@ class PriorDict(dict):
             self.conversion_function = self.default_conversion_function
 
     def evaluate_constraints(self, sample):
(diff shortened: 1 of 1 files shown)


adivijaykumar:
I am in favor of adding this, since I have fallen into this trap before. I am not sure I feel inclined to just checking against the default conversion function though. Perhaps we should also improve documentation of the PriorDict or the conversion function in both bilby and bilby_pipe to note this fact? Any other opinions?

PR #1104, opened 2026-06-26
FEAT: add precessing spin transformation

### LINKED ISSUE #698: `transform_precessing_spins` fails with array input
In GitLab by @git.ligo:samson.leong on Jul 11, 2022, 07:04

After this commit https://git.ligo.org/lscsoft/bilby/-/commit/afb2d1295fadc7960de058edc12bac0a3b39421f, which simplifies `gw.conversion.transform_precessing_spins`, the function fails with an array input like in this example:
```
args = ([1.4, 1.4, 1.4], [0.6, 0.6, 0.6], [0.3, 0.3, 0.3], [0.4, 0.4, 0.4], [0.5, 0.5, 0.5], [0.1, 0.1, 0.1], [0.2, 0.2, 0.2], [90., 90., 90.], [80., 80., 80.], [11., 11., 11.], [0.7, 0.7, 0.7])
npargs = np.array(args)
transform_precessing_spins(*npargs)
```
It returns with the errors:  
`TypeError: float() argument must be a string or a number, not 'list'`  
and  
`ValueError: setting an array element with a sequence.`

I have tried to modify slightly the function by explicitly stating the number of outputs, like this:
```
@np.vectorize
def vectorised_SimInspiralTransformPrecessingNewInitialConditions(*args):
    _1, _2, _3, _4, _5, _6, _7 = \
        SimInspiralTransformPrecessingNewInitialConditions(*args)
    return _1, _2, _3, _4, _5, _6, _7
```
and it works. My guess is without the explicit output, the `vectorise` function handles them collectively as a tuple, which would fail during array assignment.

ColmTalbot:
This is a replacement for #1044 that includes the backend changes rather than attempting to rebase.

### DIFF
--- a/bilby/gw/geometry.py
+++ b/bilby/gw/geometry.py
@@ -2,7 +2,7 @@ from plum import dispatch
 
 from .time import greenwich_mean_sidereal_time
 from ..compat.utils import array_module, promote_to_array
-from ...core.utils.constants import msun_time_si
+from ..core.utils.constants import msun_time_si
 
 
 __all__ = [
--- a/bilby/gw/geometry.py
+++ b/bilby/gw/geometry.py
@@ -1,7 +1,7 @@
 from plum import dispatch
 
 from .time import greenwich_mean_sidereal_time
-from ..compat.utils import array_module, promote_to_array
+from ..compat.utils import array_module, promote_to_array, xp_wrap
 from ..core.utils.constants import msun_time_si
 
 
@@ -378,6 +378,7 @@ def zenith_azimuth_to_theta_phi(zenith, azimuth, delta_x):
     return theta, phi
 
 
+@xp_wrap
 def transform_precessing_spins(
     theta_jn,
     phi_jl,
@@ -390,6 +391,8 @@ def transform_precessing_spins(
     mass_2,
     f_ref,
     phase,
+    *,
+    xp=None,
 ):
     """
     A direct reimplementation of
@@ -428,8 +431,6 @@ def transform_precessing_spins(
         - spin_1x, spin_1y, spin_1z: Components of spin 1
         - spin_2x, spin_2y, spin_2z: Components of spin 2
     """
-
-    xp = array_module(theta_jn)
 
     # Helper rotation functions
     def rotate_z(angle, vec):
--- a/bilby/gw/geometry.py
+++ b/bilby/gw/geometry.py
@@ -2,6 +2,7 @@ from plum import dispatch
 
 from .time import greenwich_mean_sidereal_time
 from ..compat.utils import array_module, promote_to_array
+from ...core.utils.constants import msun_time_si
 
 
 __all__ = [
@@ -472,7 +473,7 @@ def transform_precessing_spins(
     eta = mass_1 * mass_2 / (m_total * m_total)
 
     # v parameter at reference point (c=G=1 units)
-    v0 = (m_total * msun_time_si * xp.pi * f_ref) ** (1/3)
+    v0 = (m_total * msun_time_si * xp.pi * f_ref) ** (1 / 3)
 
     # Compute angular momentum magnitude using PN expressions
     # L/M = eta * v^(-1) * (1 + v^2 * L_2PN)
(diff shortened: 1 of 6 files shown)


bilby/gw/geometry.py
```diff
@@ -375,3 +375,174 @@ def zenith_azimuth_to_theta_phi(zenith, azimuth, delta_x):
     theta = xp.arccos(omega[2])
     phi = xp.arctan2(omega[1], omega[0]) % (2 * xp.pi)
     return theta, phi
+
+
+def transform_precessing_spins(
```
GregoryAshton, line 382: This is overwriting an existing function with a properly vectorised version if I have understood correctly. Typically, I'd think it would be better to use a new name. However, I can see there are tests to ensure the two agree.. so I'm inclined to say this seems like a good idea.

bilby/gw/geometry.py
```diff
@@ -375,3 +375,174 @@ def zenith_azimuth_to_theta_phi(zenith, azimuth, delta_x):
...
+    Returns
+    -------
+    tuple
+        (iota, spin_1x, spin_1y, spin_1z, spin_2x, spin_2y, spin_2z)
+        - iota: Inclination angle of L_N
+        - spin_1x, spin_1y, spin_1z: Components of spin 1
+        - spin_2x, spin_2y, spin_2z: Components of spin 2
+    """
+
+    xp = array_module(theta_jn)
```
GregoryAshton: Does array_module here just take theta_jn to figure out the array type?

GregoryAshton (commented):
Broadly looks good to me - albeit the checks are failing

bilby/gw/geometry.py
ColmTalbot: The previous version was wrapped with `np.vectorize`, so the signature hasn't changed. Both the old and new ones accept vector inputs, the only difference is that these will now be faster because it uses array broadcasting instead of `numpy.vectorize`, which is just a for loop.

bilby/gw/geometry.py
ColmTalbot: This was using an outdated version. I updated it to use the `xp_wrap` decorator, which will look at all of the input arguments.

bilby/gw/source.py
```diff
@@ -143,10 +143,16 @@ def gwsignal_binary_black_hole(frequency_array, mass_1, mass_2, luminosity_dista
...
+    try:
+        iota, spin_1x, spin_1y, spin_1z, spin_2x, spin_2y, spin_2z = bilby_to_lalsimulation_spins(
+            theta_jn=theta_jn, phi_jl=phi_jl, tilt_1=tilt_1, tilt_2=tilt_2,
+            phi_12=phi_12, a_1=a_1, a_2=a_2, mass_1=mass_1, mass_2=mass_2,
+            reference_frequency=reference_frequency, phase=phase)
+    except ZeroDivisionError:
+        if catch_waveform_errors:
+            return None
+        else:
+            raise
```
adivijaykumar, line 155: Would it be possible to abstract this out to a separate function? It is repeated three times in the same file.

bilby/gw/geometry.py
adivijaykumar, line 18: I think `transform_precessing_spins` should be added here.

adivijaykumar (changes requested):
I like this change! I am worried about regressions that it introduces though. Quick checks run on my laptop show that this could make up ~10% of a single phenomPv2 call, as opposed to <1% before. It might be an okay tradeoff in anticipation of future developments. Any thoughts @ColmTalbot?

Also, I was a bit surprised to find the function in geometry.py. I have always assumed that it pertains to detector geometry and location. Not totally opposed to this being in geometry.py, but I still think it fits better in conversion.py.

adivijaykumar:
Thanks @ColmTalbot! This is encouraging. I had performed the test by making waveform calls over a bunch of parameters, but I agree that one of the tutorials is probably the best to benchmark this. I also found really massive improvements with `jax` calls over a large number of parameters (even on CPU), so I am in favor of adding this in!

PR #1106, opened 2026-06-29
BUG: fix ROQ likelihood incompatibility with numpy 2.4

mj-will:
The ROQ likelihood had a small bug related to how the HDF5 bases were being loaded that caused an error with numpy 2.4.

I've put in a small fix and make `roq_params` a property so it's shape is checked.

### DIFF
--- a/bilby/gw/likelihood/roq.py
+++ b/bilby/gw/likelihood/roq.py
@@ -90,14 +90,22 @@ class ROQGravitationalWaveTransient(GravitationalWaveTransient):
 
     @roq_params.setter
     def roq_params(self, roq_params):
-        if (
-            isinstance(roq_params, np.ndarray)
-            and roq_params.dtype.names is not None
-        ):
+        if roq_params is not None:
+            if not isinstance(roq_params, np.ndarray) or roq_params.dtype.names is None:
+                raise TypeError(
+                    "roq_params must be None or a scalar structured numpy array"
+                )
             if roq_params.ndim != 0:
                 raise ValueError(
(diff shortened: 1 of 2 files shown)


bilby/gw/likelihood/roq.py
```diff
@@ -84,6 +84,23 @@ class ROQGravitationalWaveTransient(GravitationalWaveTransient):
...
 
     """
+    @property
+    def roq_params(self):
+        return self._roq_params
+
+    @roq_params.setter
+    def roq_params(self, roq_params):
+        if (
+            isinstance(roq_params, np.ndarray)
```
ColmTalbot: Is this leaving space for some other input type?
Will this work with a dictionary or dataframe?

ColmTalbot (commented):
I'm happy with this change as long as the tests pass, just one question.

bilby/gw/likelihood/roq.py
mj-will: We have to also support the `roq_params = None` case, see https://github.com/mj-will/bilby/blob/numpy2p4-support/bilby/gw/likelihood/roq.py#L132.

I could remove the clause after the `and` statement. It might make sense to replace it with a check for the three mandatory keys.

L131 would suggest we don't support dictionaries or dataframes.

mj-will: I've made the check more explicit.

Also, it seems the the test failure is unrelated.

GregoryAshton (approved):
LGTM

adivijaykumar (approved):

PR #1108, opened 2026-06-30
Make Prior.__repr__ round-trippable for numpy scalar parameters

lorenzopompili00:
With numpy 2.0+, `repr()` of numpy scalars carry an `np.` prefix (e.g. `repr(np.float64(0.0))` is `"np.float64(0.0)"`). This goes into `Prior.__repr__`, and bilby's parser `Prior._parse_argument_string` then fails to re-parse it: it treats`np` as a module and calls `import_module("np")` which fails.

Can be reproduced by running
```python
import numpy as np
from bilby.core.prior import Uniform, PriorDict

p = Uniform(minimum=np.float64(0.0), maximum=np.float64(1.0), name="x")
print(repr(p))          # Uniform(minimum=np.float64(0.0), maximum=np.float64(1.0), ...)
PriorDict({"x": repr(p)})   # ModuleNotFoundError: No module named 'np'
```

This fixes it by converting numpy scalars to native Python types in `Prior.__repr__`

### DIFF
--- a/bilby/core/prior/base.py
+++ b/bilby/core/prior/base.py
@@ -253,7 +253,19 @@ class Prior(object):
         prior_name = self.__class__.__name__
         prior_module = self.__class__.__module__
         instantiation_dict = self.get_instantiation_dict()
-        args = ', '.join([f'{key}={repr(instantiation_dict[key])}' for key in instantiation_dict])
+
+        def _native(value):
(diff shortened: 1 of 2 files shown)


ColmTalbot (approved):
Thanks @lorenzopompili00! I hadn't come across numpy.generic.

adivijaykumar (approved):

PR #1116, opened 2026-07-21
BUG: Validate PriorDict assignments

### LINKED ISSUE #1111: BUG: PriorDict silently drops invalid values added after initialization
Note: the bug report below was prepared by AI under my supervision.

**Describe the bug**

`PriorDict` validates entries supplied during initialization, but apparently does not apply the same validation when entries are subsequently added using item assignment.

A zero-dimensional NumPy array is rejected when passed to the `PriorDict` constructor:

```python
priors = PriorDict({
    "x": Uniform(0, 1, name="x"),
    "Om0": np.array(0.30966),
})
```

This raises a clear `TypeError`.

However, the same value can be added after the `PriorDict` has been instantiated:

```python
priors = PriorDict({"x": Uniform(0, 1, name="x")})
priors["Om0"] = np.array(0.30966)
```

The invalid entry then appears in the prior dictionary, but is silently omitted by `PriorDict.sample()`.

This creates inconsistent behavior depending on whether an entry is supplied during initialization or assigned afterward. More importantly, the post-initialization path fails silently: the parameter remains visible in `priors.keys()` but is absent from the sampled parameter dictionary without any error or warning.

**To Reproduce**

```python
import numpy as np
from bilby.core.prior import PriorDict, Uniform
```

First, supplying a zero-dimensional NumPy array during initialization raises an error:

```python
priors = PriorDict({
    "x": Uniform(0, 1, name="x"),
    "H0": np.float64(67.66),
    "Om0": np.array(0.30966),
    "w0": np.array(-1.0),
})
```

This produces:

```text
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/andreasb/.conda/envs/gwpop_jax_gpu/lib/python3.11/site-packages/bilby/core/prior/dict.py", line 37, in __init__
    self.from_dictionary(dictionary)
  File "/home/andreasb/.conda/envs/gwpop_jax_gpu/lib/python3.11/site-packages/bilby/core/prior/dict.py", line 305, in from_dictionary
    raise TypeError(
TypeError: Unable to parse prior, bad entry: Om0 = 0.30966 of type <class 'numpy.ndarray'>
```

However, adding the same entry after initialization does not raise an error:

```python
priors = PriorDict({
    "x": Uniform(0, 1, name="x"),
})

priors["Om0"] = np.array(0.30966)

print(priors)
print(priors.keys())
```

The output is:

```text
{
    'x': Uniform(
        minimum=0,
        maximum=1,
        name='x',
        latex_label='x',
        unit=None,
        boundary=None
    ),
    'Om0': array(0.30966)
}

dict_keys(['x', 'Om0'])
```

Calling `sample()` then silently omits `Om0`:

```python
sample = priors.sample()

print(sample)
print(sample.keys())
```

For example:

```text
{'x': 0.4084391218257899}

dict_keys(['x'])
```

Thus, NumPy scalar values such as `np.float64` are retained, while zero-dimensional `numpy.ndarray` values are silently omitted.

**Error message**

When the zero-dimensional array is supplied during initialization, the full error is:

```text
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/andreasb/.conda/envs/gwpop_jax_gpu/lib/python3.11/site-packages/bilby/core/prior/dict.py", line 37, in __init__
    self.from_dictionary(dictionary)
  File "/home/andreasb/.conda/envs/gwpop_jax_gpu/lib/python3.11/site-packages/bilby/core/prior/dict.py", line 305, in from_dictionary
    raise TypeError(
TypeError: Unable to parse prior, bad entry: Om0 = 0.30966 of type <class 'numpy.ndarray'>
```

When the same entry is added after initialization using:

```python
priors["Om0"] = np.array(0.30966)
```

there is no error message or warning. The entry is accepted into the dictionary and is later silently omitted by `PriorDict.sample()`.

**Expected behavior**

The behavior should be consistent regardless of whether an entry is provided to the constructor or assigned after initialization.

There are at least two reasonable behaviors:

1. Zero-dimensional NumPy arrays are considered invalid fixed prior values. In that case, assigning one after initialization should raise the same `TypeError` as supplying one to the constructor.

2. Zero-dimensional NumPy arrays are considered scalar fixed values. In that case, Bilby should retain them during sampling, perhaps after converting them to NumPy or Python scalars using `.item()`.

Under no circumstances should a key remain visible in the `PriorDict` but then be silently absent from the output of `PriorDict.sample()`.

At minimum, `PriorDict.sample()` should raise an error or warning when it encounters an unsupported entry type.

**Suggested solution (optional)**

Apply the same validation used by `PriorDict.from_dictionary()` whenever a new value is assigned through `PriorDict.__setitem__()`.

For example, assignment such as:

```python
priors["Om0"] = np.array(0.30966)
```

could raise the same error as constructor initialization:

```text
TypeError: Unable to parse prior, bad entry: Om0 = 0.30966 of type <class 'numpy.ndarray'>
```

Alternatively, Bilby could explicitly support zero-dimensional NumPy arrays as fixed scalar values:

```python
if isinstance(value, np.ndarray) and value.ndim == 0:
    value = value.item()
```

It may also be useful for `PriorDict.sample()` to check that every entry has either been sampled or copied into the output. If an entry is skipped because its type is unsupported, it should raise an informative error rather than silently dropping the parameter.

A user-side workaround is:

```python
priors["Om0"] = np.asarray(Om0).item()
priors["w0"] = np.asarray(w0).item()
```

but the current behavior is difficult to detect because assignment succeeds and the key remains present in the prior dictionary.

**Environment (please complete the following information):**

* Bilby version: `[2.8.0]`
* Installation method [e.g. conda, pip, source]: `[conda]`
* Python version: `3.11`

**Additional context**

This arose in a larger prior dictionary containing fixed cosmological parameters:

```python
{
    ...
    "H0": np.float64(67.66),
    "Om0": np.array(0.30966),
    "w0": np.array(-1.0),
}
```

The entries were added after the `PriorDict` had already been instantiated. All three keys appeared in `fast_priors.keys()`, but only `H0` appeared in the dictionary returned by `fast_priors.sample()`.

The relevant types and shapes are:

```python
type(fast_priors["H0"])
# numpy.float64

type(fast_priors["Om0"])
# numpy.ndarray

fast_priors["Om0"].shape
# ()

type(fast_priors["w0"])
# numpy.ndarray

fast_priors["w0"].shape
# ()
```

Therefore, the primary issue is not simply that zero-dimensional arrays are unsupported. The issue is that they are rejected during construction but accepted through later assignment, after which they are silently discarded during sampling.

Sanjays2402:
Closes #1111

Reject unsupported values assigned to `PriorDict` with the same `TypeError` used during constructor parsing. This prevents invalid entries such as zero-dimensional NumPy arrays from remaining visible in the dictionary while silently disappearing from `sample()` output.

Tested with the full `test/core/prior/dict_test.py` suite (37 passed), including a regression that fails on the original source.

### DIFF
--- a/bilby/core/prior/dict.py
+++ b/bilby/core/prior/dict.py
@@ -60,6 +60,14 @@ class PriorDict(dict):
     def __hash__(self):
         return hash(str(self))
 
+    def __setitem__(self, key, value):
+        if not isinstance(value, (Prior, int, float, str, dict)):
+            raise TypeError(
+                "Unable to parse prior, bad entry: {} "
+                "= {} of type {}".format(key, value, type(value))
+            )
+        super().__setitem__(key, value)
+
     @xp_wrap
     def evaluate_constraints(self, sample, *, xp=None):
         out_sample = self.conversion_function(sample)
--- a/test/core/prior/dict_test.py
+++ b/test/core/prior/dict_test.py
@@ -79,6 +79,11 @@ class TestPriorDict(unittest.TestCase):
     def test_prior_set_has_expected_priors(self):
         self.assertDictEqual(self.priors, dict(self.prior_set_from_dict))
 
+    def test_assignment_rejects_invalid_prior_value(self):
+        expected_error = "Unable to parse prior, bad entry"
+        with self.assertRaisesRegex(TypeError, expected_error):
+            self.prior_set_from_dict["Om0"] = np.array(0.30966)
+
     def test_read_from_file(self):
         expected = dict(
             mass_1=bilby.core.prior.Constraint(


bilby/core/prior/dict.py
```diff
@@ -60,6 +60,14 @@ def __init__(self, dictionary=None, filename=None, conversion_function=None):
     def __hash__(self):
         return hash(str(self))
 
+    def __setitem__(self, key, value):
+        if not isinstance(value, (Prior, int, float, str, dict)):
```
mj-will, line 64: I'm not sure why `str` is allowed here, can you help me understand?

mj-will (commented):
Hi @Sanjays2402, thanks for implementing this fix.

Just a small question from me.

Sanjays2402:
Good question, @mj-will — `str` is allowed because the constructor parsing already accepts strings: every value read from a `.prior` file arrives as a string (e.g. `foo = Uniform(minimum=0, maximum=1)`), and `from_dictionary` parses those into `Prior` objects (bare numbers like `"1"` become `DeltaFunction`s). The `__setitem__` check is meant to mirror exactly the set of types the constructor accepts and parses, rather than introduce a stricter contract — otherwise something the constructor itself handles would start being rejected on assignment.

mj-will:
> Good question, @mj-will — `str` is allowed because the constructor parsing already accepts strings: every value read from a `.prior` file arrives as a string (e.g. `foo = Uniform(minimum=0, maximum=1)`), and `from_dictionary` parses those into `Prior` objects (bare numbers like `"1"` become `DeltaFunction`s). The `__setitem__` check is meant to mirror exactly the set of types the constructor accepts and parses, rather than introduce a stricter contract — otherwise something the constructor itself handles would start being rejected on assignment.

Thanks @Sanjays2402. I see why `str` is there. However, I think we might need different logic here since the string inputs in the constructor a parsed in `from_dictionary` but `__setitem__` doesn't parse them. That means that, with the current change, a string would remain a string, leading to an invalid entry in `PriorDict`. I think same is also true of `dict`. This isn't an issue with floats/ints because the `convert_floats_to_delta_functions()` will convert those when e.g. `sample is called`.

We could consider changing `__setitem__` so it also parses strings and dictionaries, but that would require refactoring more of the code. So I suggest we leave that for another PR and change the check here to `isinstance(value, (Prior, int, float))`. What do you think?

Sanjays2402:
Good point about the unparsed strings — I tried the narrowing to isinstance(value, (Prior, int, float)) locally, and it breaks two existing tests that deliberately rely on string assignment surviving __setitem__: test_convert_floats_to_delta_functions assigns "unconvertable" and expects it to stay a string after conversion, and test_sample_subset_correct_size_when_non_priors_in_dict assigns "not_a_prior" and expects sample_subset to skip it. So narrowing would also mean changing the intent of those tests. Happy to go either way — narrow the check and update those two tests to match, or keep the current check and do the parse-strings-in-__setitem__ refactor as the follow-up. Which would you prefer?

mj-will:
Thanks for weighing in Colm, I agree, I think we should scrap those tests.

Sanjays2402:
@mj-will @ColmTalbot — done in c85bb3b0, per your call. Narrowed __setitem__ to (Prior, int, float, dict), and removed test_convert_floats_to_delta_functions and test_sample_subset_correct_size_when_non_priors_in_dict entirely since both relied on string assignment surviving. Parseable string specs (e.g. "PowerLaw(...)") still convert via from_dictionary before the check, and test_from_dict_with_string passes. Dict suite is 34 passed locally (the one failure is just healpy missing in my env, same on the base commit). Let me know if you'd rather keep the float-conversion assertions somewhere.

test/core/prior/dict_test.py
```diff
@@ -181,25 +186,6 @@ def test_from_dict_with_string(self):
...
-            speed=bilby.core.prior.PowerLaw(
-                name="b", alpha=3, minimum=1, maximum=2, unit="m/s", boundary=None
-            ),
-            length=bilby.core.prior.DeltaFunction(name="c", peak=42, unit="m"),
-            d=bilby.core.prior.DeltaFunction(peak=5),
-            e=bilby.core.prior.DeltaFunction(peak=7.3),
-            f="unconvertable",
-        )
-        self.assertDictEqual(expected, self.prior_set_from_dict)
-
```
mj-will, line 202: Can we add this test back, but just remove `self.prior_set_from_dict["f"] = "unconvertable"` and the related lines below.

bilby/core/prior/dict.py
```diff
@@ -60,6 +60,14 @@ def __init__(self, dictionary=None, filename=None, conversion_function=None):
     def __hash__(self):
         return hash(str(self))
 
+    def __setitem__(self, key, value):
+        if not isinstance(value, (Prior, int, float, dict)):
```
mj-will: I think we should remove `dict` here since

```
priors = PriorDict({"x": Uniform(0, 1)})
priors["y"] = {"a": 42}
priors.sample()
```
doesn't work.

mj-will (changes requested):
Thanks @Sanjays2402, a couple of comments from on further changes I think we should make.

Sanjays2402:
Done both, @mj-will: removed `dict` from the `__setitem__` check, so `priors["y"] = {"a": 42}` now raises the same TypeError as the other bad values. The constructor path is unaffected — `from_dictionary` parses dicts itself and assigns via `update()`, which doesn't go through `__setitem__`. Also restored `test_convert_floats_to_delta_functions` minus the unconvertable-string lines.

mj-will (approved):
Great, thanks @Sanjays2402, LGTM now!

ColmTalbot (approved):

PR #1117, opened 2026-07-27
Fix passing of proposals and bounds in the dynesty 3 api

### LINKED ISSUE #1065: [BUG]: The dynesty 3 api passes neither the bound-kwarg nor the proposals-kwarg
In the recent updates to the dynesty api to support dynesty 3, both the `bound` and `proposals` options have no effect when selecting any of the bilby-implemented samplers. As a result, only differential evolution can be used as a proposal scheme, and the volumetric proposals would be very inefficient as they would always use the identity matrix as their proposal axes.

For the proposal method, there is a one-line fix in `dynesty.samler_init_kwargs`.

For the bounding, it is a bit more subtle because with dynesty3, all dynesty bounding options should just work for the volumetric proposals.
For differential evolution, they would have no effect anyway. So it would be ok to completely remove any reference to the 'live' bounding and rephrase the documentation a bit.

I have implemented both these based on  https://github.com/bilby-dev/bilby/pull/1049 dropping dynesty<3 support [here](https://github.com/JasperMartins/bilby/tree/remove-old-dynesty-fixes).

JasperMartins:
This MR fixes the issues discussed in https://github.com/bilby-dev/bilby/issues/1065.

Closes https://github.com/bilby-dev/bilby/issues/1065

Moved from https://github.com/ColmTalbot/bilby/pull/4

### DIFF
--- a/bilby/core/sampler/dynesty.py
+++ b/bilby/core/sampler/dynesty.py
@@ -266,6 +266,7 @@ class Dynesty(NestedSampler):
             periodic=self.kwargs.get("periodic", None),
             reflective=self.kwargs.get("reflective", None),
             maxmcmc=self.maxmcmc,
+            proposals=self.proposals,
         )
 
         if kwargs["sample"] == "act-walk":
--- a/bilby/core/sampler/dynesty.py
+++ b/bilby/core/sampler/dynesty.py
@@ -266,8 +266,10 @@ class Dynesty(NestedSampler):
             periodic=self.kwargs.get("periodic", None),
             reflective=self.kwargs.get("reflective", None),
             maxmcmc=self.maxmcmc,
-            proposals=self.proposals,
         )
+
+        if self.proposals:
+            internal_kwargs["proposals"] = self.proposals
 
         if kwargs["sample"] == "act-walk":
             internal_kwargs["nact"] = self.nact
--- a/bilby/core/sampler/dynesty.py
+++ b/bilby/core/sampler/dynesty.py
@@ -128,8 +128,9 @@ class Dynesty(NestedSampler):
     nlive: int, (1000)
         The number of live points, note this can also equivalently be given as
         one of [nlive, nlives, n_live_points, npoints]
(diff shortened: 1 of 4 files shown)


JasperMartins:
I'm unsure about the failing jax-backend test. I don't think it is related to this MR 🤔

ColmTalbot (approved):
This looks good to me. Once we get the CI fixed, I think we can get this merged in.

adivijaykumar (approved):

bilby/core/sampler/dynesty.py
```diff
@@ -290,24 +291,14 @@ def sampler_init_kwargs(self):
         elif kwargs["sample"] == "rwalk":
             internal_kwargs["nact"] = self.nact
             internal_sampler = dynesty_utils.AcceptanceTrackingRWalk(**internal_kwargs)
-            bound = "none"
             logger.info(
                 f"Using the bilby-implemented ensemble rwalk sampling method with ACT "
                 f"estimated chain length. An average of {2 * internal_sampler.nact} "
                 f"steps will be accepted up to chain length {internal_sampler.maxmcmc}."
             )
-        elif kwargs["bound"] == "live":
```
ColmTalbot, line 299: I think we need to reinstate this clause.

ColmTalbot (changes requested):
I think, actually, this has broken the default behaviour when not using the live-point sampling.
The docs job timing out is because the default bound for the uniform sampling is now no bound, which is horribly inefficient.

bilby/core/sampler/dynesty.py
JasperMartins: Alternatively, `bound = "none"` could be dropped entirely as a default keyword argument for the dynesty sampler. That would make "live-multi" (or whatever the upstream default bounding option is if it ever changes) the default option. If it is preferred that the behavior of `"live"` stays the default, I'll reinstate the clause, otherwise I'll push a fix that removes the default option entirely.

PR #1118, opened 2026-07-27
ENH: Adding Mixture-Class for superposition of priors

Hen42rik:
This is a new PR for the same suggested addition of #1077, but addressing the following additional aspects:
1. as requested by @fgittins , the Prior class is renamed to `Mixture`
2. as requested by @fgittins , it is refactored to inherit most methods from the `Interped` class
3. the interface is adjusted to respect the new xp_wrap decorators

The original

### DIFF
--- a/bilby/core/prior/base.py
+++ b/bilby/core/prior/base.py
@@ -390,19 +390,7 @@ class Prior(object):
     @classmethod
     def _split_repr(cls, string):
         subclass_args = infer_args_from_method(cls.__init__)
-        args = string.split(',')
-        remove = list()
-        for ii, key in enumerate(args):
-            for paren_pair in ['()', '{}', '[]']:
-                if paren_pair[0] in key:
-                    jj = ii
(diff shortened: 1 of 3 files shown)


Hen42rik:
I would consider this now complete, but happy to consider further remarks @fgittins  @mj-will 
P.S: Sorry for the delay, holiday

GregoryAshton:
@Hen42rik this also seems to be doing some rewriting of `_split_repr` which I think was used to ensure the written repr could be used to reconstruct the prior. I do see extensions to the tests, but I'm trying to figure out if there are any edge cases we need to add. Just leaving this as a comment to discuss

Hen42rik:
@GregoryAshton I agree that some pitfalls might be hidden here. I have spent some extra time to make sure some of the previously skipped prior tests are now performed, too, but I am not able to vow for this to not break something. I think someone more familiar with these edge cases would probably be more suitable to suggest additional required tests.

PR #1126, opened 2026-08-10
BLD: Switch CI tests to use pixi for environments

### LINKED ISSUE #1048: Remove lal install in CI
After https://github.com/bilby-dev/bilby/pull/1043, we have explicit reinstalls of lal in the CI jobs, we should remove these with a more permanent fix.

### LINKED ISSUE #1123: CI: ensure a version of the CI runs without jax

ColmTalbot:
This is an attempt to remove our need to generate containers in scheduled jobs and have them occasionally just fail because of upstream changes and instead use [pixi](https://pixi.prefix.dev/latest/) (conda+pypi+rust) with a lock file.
The lockfile can be modified on a schedule using pull requests so we know updating our test environment won't break things.


.github/workflows/unit-tests.yml
```diff
@@ -109,20 +173,25 @@ jobs:
...
+      id: test-data-cache
+      uses: actions/cache/restore@v6
+      with:
+        path: .roq-data
+        key: roq-data-v1
+
+    - uses: prefix-dev/setup-pixi@v0.10.0
+      with:
+        environments: ${{ matrix.python.environment }}
+        cache: true
```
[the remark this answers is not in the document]
ColmTalbot: This is working...

.github/workflows/unit-tests.yml
```diff
@@ -20,51 +20,79 @@ concurrency:
...
+        cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
+        locked: true
+
+    - name: Run tests
       run: |
-        pytest test/integration/sampler_run_test.py --durations 10 -v        
+        pixi run -e ${{ matrix.python.environment }} \
+          pytest . --cov=bilby --cov-branch \
+          --durations 10 -ra --color yes --cov-report=xml --junitxml=pytest.xml
```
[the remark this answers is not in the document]
ColmTalbot: Another swing and miss, these tests are exluded by the configuration.

pixi.toml
```diff
@@ -0,0 +1,110 @@
...
+sphinx = "<9"
+sphinx-design = "*"
+sphinx-sitemap = "*"
+sphinx_rtd_theme = "*"
+
+[feature.doc.pypi-dependencies]
+autodoc = "*"
+myst_parser = "*"
+nessai-bilby = "*"
+nessai-gw = "*"
```
mj-will: This shouldn't be needed


```suggestion

```

pixi.toml
```diff
@@ -0,0 +1,110 @@
...
+liblal = "!=7.7.0"
+ligo-gracedb = "*"
+pyfftw = "*"
+pyseobnr = "*"
+python-lal = "!=7.7.0"
+python-lalframe = "*"
+python-lalsimulation = "*"
+
+[feature.gw.pypi-dependencies]
+bilby-cython = ">=0.5.4"
```
mj-will, line 65: Why use pypi for bilby-cython?

containers/environment.yml
```diff
@@ -1,71 +0,0 @@
...
-  - sphinx<9 # needed for sphinx_rtd_theme<3.1
-  - numpydoc
-  - nbsphinx
-  - sphinx_rtd_theme
-  - sphinx-design
-  - dynesty>=3.0.0
-  - emcee
-  - pytorch
-  - pytorch-cpu
-  - numba>0.53.1
```
mj-will, line 47: numba doesn't seem to be in the new env, is that intentional?

containers/environment.yml
```diff
@@ -1,71 +0,0 @@
-# This is a template yaml file for the test image
-# The python version should be added before creating the env
-channels:
-  - conda-forge
-dependencies:
-  - pip
-  - setuptools
-  - setuptools_scm
```
mj-will, line 8: I don't think setuptools_scm is installed anywhere, not sure if need it in the tests though

pixi.toml
```diff
@@ -0,0 +1,110 @@
...
+[environments]
+documentation = ["py313", "gw", "doc", "test", "jax", "torch"]
+jax = ["py313", "gw", "test", "array_api", "jax"]
+latest-dynesty = ["py313", "gw", "test", "dynesty"]
+numpy = ["py313", "gw", "test", "array_api"]
+py311 = ["py311", "gw", "test"]
+py312 = ["py312", "gw", "test"]
+py313 = ["py313", "gw", "test"]
+py313-base = ["py313"]
+torch = ["py313", "gw", "test", "array_api", "torch"]
```
mj-will, line 26: In future, would it make sense to have the GW test suite run seperately?

containers/environment.yml
```diff
@@ -1,71 +0,0 @@
...
-  - configargparse
-  - future
-  - dill
-  - black
-  - pytest-cov
-  - pytest-requires
-  - pytest-rerunfailures
-  - arviz
-  - parameterized
-  - scikit-image
```
mj-will, line 25: scikit-image is not longer present, I think this is likely okay though.

pixi.toml
```diff
@@ -0,0 +1,110 @@
...
+pandoc = "*"
+sphinx = "<9"
+sphinx-design = "*"
+sphinx-sitemap = "*"
+sphinx_rtd_theme = "*"
+
+[feature.doc.pypi-dependencies]
+autodoc = "*"
+myst_parser = "*"
+nessai-bilby = "*"
```
mj-will, line 45: nessai-bilby can be installed via conda

test/core/prior/base_test.py
```diff
@@ -189,9 +189,8 @@ def rescale(self, val):
                     return val * 2
 
             prior = CustomPriorWithoutXp(name="custom_prior")
-            import jax.numpy as jnp
-            rescaled = prior.rescale(jnp.array([0.1, 0.2, 3]))
-            self.assertEqual(aac.get_namespace(rescaled), jnp)
+            rescaled = prior.rescale(np.array([0.1, 0.2, 3]))
+            self.assertEqual(aac.get_namespace(rescaled), aac.numpy)
```
mj-will, line 212: Why was this changed needed?

mj-will (commented):
Overall, LGTM. Just a couple of minor comments and questions.

pixi.toml
ColmTalbot: There was an issue with resolving it for some platforms/versions and it was easier to just depend on the pypi version.
We can try to address this in a follow up.

test/core/prior/base_test.py
ColmTalbot: This is because jax is no longer in the base testing environment, so would have been flagged by https://github.com/bilby-dev/bilby/issues/1123.

pixi.toml
ColmTalbot: Yup, I'd rather defer to later, and this should make it much more straightforward.

containers/environment.yml
ColmTalbot: We should probably retain this, I suspect it will upset the pages at least.

containers/environment.yml
ColmTalbot: Oh, this is needed to generate the logo, but I guess we don't do that in CI.

containers/environment.yml
ColmTalbot: It isn't explicitly pinned, but it is in the env. I think this pin was to handle an undeclared upstream requirement. I tried to remove transitive pins like those.

pixi.toml
ColmTalbot: To clarify, this isn't needed for our test suite?

pixi.toml
ColmTalbot: This introduces a circular dependency, which made it very unhappy.

```python
$ pixi lock
Error:   × failed to solve the pypi requirements of environment 'documentation' for platform 'osx-arm64'
  ├─▶ failed to resolve pypi dependencies
  ╰─▶ Because there is no version of bilby==2.8.2 and you require bilby==2.8.2, we can conclude that your requirements are unsatisfiable.
  help: The following PyPI packages have been pinned by the conda solve, and this version may be causing a conflict:
        bilby==2.8.2
        See https://pixi.sh/latest/concepts/conda_pypi/#pinned-package-conflicts for more information.
```

pixi.toml
mj-will: Yeah, that's right.

pixi.toml
mj-will: Okay, I've created an issue for this: https://github.com/bilby-dev/bilby/issues/1135

containers/environment.yml
mj-will: I see this has been added.

pixi.toml
mj-will: Weird, I dug into this at it seems that the issue is that `nessai-bilby` depends on `bilby` and if one installs that from `conda` then it pins the versions to `bilby==2.8.2`. The editable `pip` install of `bilby` for the tests then fails since it doesn't provided 2.8.2.

From some digging there may be ways around this but I'm not familiar enough with `pixi` to know if they make sense. For example:

```
[workspace.conda-pypi-map]
conda-forge = { mapping = { bilby = false } }
```

apparently tells `pixi` to not pass the `conda` version to `pip`.

adivijaykumar (approved):
Looks good to me!

pixi.toml
ColmTalbot: Can we defer this to a follow-up PR?

mj-will (approved):

PR #1129, opened 2026-08-24
BUG: Function Parameter Inference

mick-wright:
Currently, the [`infer_parameters_from_function`](https://github.com/bilby-dev/bilby/blob/0e77aa01f2fb5f2432edcd76cd764175aac4b7a7/bilby/core/utils/introspection.py#L5) function fails for functions coming from python C-extensions since these functions are classed as `builtin_function_or_method`.

To fix this, this PR simply expands the `isinstance` check to inclue `types.BuiltinFunctionType` and `types.BuiltinMethodType` which allows these functions to be included, assuming that they have signature information for inspect to find.

### DIFF
--- a/bilby/core/utils/introspection.py
+++ b/bilby/core/utils/introspection.py
@@ -102,7 +102,7 @@ def infer_args_from_function_except_n_args(func, n=1):
         inspect.signature(func).parameters.values()
         if not parameter.kind in (
             inspect.Parameter.VAR_POSITIONAL,
-            inspect.Parameter.VAR_KEYWROD
+            inspect.Parameter.VAR_KEYWORD
         )
     ]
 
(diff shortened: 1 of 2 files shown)


mj-will (changes requested):
Thanks for fixing this @mick-wright, the changes LGTM.

Could you also add a simple test for this function seeing as we're making changes to it?

bilby/core/utils/introspection.py
```diff
@@ -31,13 +30,12 @@ def infer_parameters_from_function(func):
     This allows the reference to the instance (conventionally named `self`)
     to be removed.
```
mj-will: I think we should probably remove the mention of the `type` here.

mj-will (changes requested):
Looks good to me, just a small comment on the doc-string.

bilby/core/utils/introspection.py
mick-wright: The docstring has been adjusted accordingly.

mj-will (approved):

mattpitkin (approved):

PR #1138, opened 2026-08-27
BUG(dynesty): check parameter keys match when resuming

### LINKED ISSUE #956: Checkpointing is scrambling parameter labels.
I've run into a strange error where checkpointing is scrambling the parameter labels. Trace plots are produced correctly and the final result file looks sensible as long as bilby is not loading a checkpoint during its run.  This is NOT happening all the time but only with this particular model / parameter labels. My hunch is that my parameter names are probably breaking some hidden rule

In one of the runs where I did load from a checkpoint, most of the posterior samples of this parameter are outside its prior range. And the even more bizarre thing is that the every time I read in the checkpoint file, I get a new random permutation of the label-sample association. 

```
(Pdb) priors["sigma_p_0"]
Uniform(minimum=0.01, maximum=1.0, name='sigma_p_0', latex_label='$\\sigma^0_{p}$', unit=None, boundary=None)
(Pdb) result.posterior['sigma_p_0'].max()
4.862758644802147
``` 
For reference, here are the list of parameter labels I'm using. I'm using dynesty. 

 ```
(Pdb) priors.keys()
dict_keys(['mb_0', 'beta_0', 'alpha', 'sigpp_2', 'lam_0', 'lamb_0', 'sigpp_1', 'lam_1', 'delta_m_2', 'mpp_1', 'mmax', 'sigma_eff_0', 'beta_1', 'sigma_p_0', 'mu_p_0', 'mpp_2', 'mlow_2', 'delta_m_1', 'mu_eff_0', 'mlow_1'])
```

mj-will:
When resuming dynesty runs, if `search_parameter_keys` list can change order since it relies on the ordering from a dictionary. As reported in https://github.com/bilby-dev/bilby/issues/956, this can lead to invalid resume states.

This PR adds the keys to checkpoint metadata so I can be checked when resuming.

Closes https://github.com/bilby-dev/bilby/issues/956

### DIFF
--- a/bilby/core/sampler/dynesty.py
+++ b/bilby/core/sampler/dynesty.py
@@ -18,7 +18,13 @@ from ..utils import (
 )
 from ..utils.plotting import _close_new_figures
 from . import dynesty_utils
-from .base_sampler import NestedSampler, Sampler, _SamplingContainer, signal_wrapper
+from .base_sampler import (
+    NestedSampler,
+    ResumeError,
+    Sampler,
+    _SamplingContainer,
+    signal_wrapper,
+)
 
 
 def _set_sampling_kwargs(args):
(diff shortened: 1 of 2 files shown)


GregoryAshton (approved):
This looks good to me

adivijaykumar (approved):

PR #1140, opened 2026-08-28
MAINT: change default logging configuration

### LINKED ISSUE #834: Logger configuration
Currently, the logger in bilby adds handlers by default (see [here](https://github.com/bilby-dev/bilby/blob/133174a6a160889d95e97a03f2c8f15d4b3479b5/bilby/core/utils/__init__.py#L17)). This goes against the [recommend practice](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library):

> Note: It is strongly advised that you do not add any handlers other than [NullHandler](https://docs.python.org/3/library/logging.handlers.html#logging.NullHandler) to your library’s loggers. This is because the configuration of handlers is the prerogative of the application developer who uses your library. The application developer knows their target audience and what handlers are most appropriate for their application: if you add handlers ‘under the hood’, you might well interfere with their ability to carry out unit tests and deliver logs which suit their requirements.

I think we should consider changing this, what do people think?

mj-will:
Address https://github.com/bilby-dev/bilby/issues/834 by changing the logging configuration to only add the null logger by default.

This follows python best practices and makes it easier for downstream users to handle bilby logging.

Replaces https://github.com/bilby-dev/bilby/pull/1072 since it was closed by the author.

### DIFF
--- a/bilby/__init__.py
+++ b/bilby/__init__.py
@@ -16,7 +16,11 @@ https://bilby-dev.github.io/bilby/installation.html.
 """
 
 
+import logging
 import sys
+
+# Avoid configuring logging for applications that import bilby.
+logging.getLogger(__name__).addHandler(logging.NullHandler())
 
 from . import core, gw, hyper
 
--- a/bilby/core/utils/__init__.py
+++ b/bilby/core/utils/__init__.py
@@ -18,5 +18,3 @@ from .series import *
 
 #  Instantiate the default argument parser at runtime
 command_line_args, command_line_parser = set_up_command_line_arguments()
-#  Instantiate the default logging
-setup_logger(print_version=False, log_level=command_line_args.log_level)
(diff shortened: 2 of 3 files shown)


bilby/__init__.py
```diff
@@ -16,8 +16,12 @@
 """
 
 
+import logging
 import sys
 
+# Avoid configuring logging for applications that import bilby.
```
ColmTalbot, line 19: Is there a reason to do this here and not in logging.py?

mj-will: This ensures the logger is always configured before any of the submodules are imported and could produce a logging statement.

For example, if it were in `log.py` and a log statement were produced by one of the submodules before `log.py` was imported, it wouldn't be caught by the `NullHandler`.

ColmTalbot (approved):

adivijaykumar (approved):

PR #1141, opened 2026-08-28
MAINT: remove python<3 check

mj-will:
I noticed this check for Python < 3 still exists. Given Python 2.0 is long gone, I thought it best we removed it.


mattpitkin (approved):

ColmTalbot (approved):
Agree we should remove this, but I’ve been actively ignoring this for a long time.

PR #1142, opened 2026-09-02
Add workflow to update pixi lock files

ColmTalbot:
This workflow updates lock files weekly using pixi and creates a pull request with the changes. It can also be manually triggered.

### DIFF
--- a/.github/workflows/update-lock-files.yml
+++ b/.github/workflows/update-lock-files.yml
@@ -1,0 +1,36 @@
+name: Update lock files
+
+permissions: 
+  contents: write
+  pull-requests: write
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: 0 5 * * 1
+
+jobs:
+  pixi-update:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


adivijaykumar (approved):

mj-will (approved):

.github/workflows/update-lock-files.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Update lock files
+        run: |
+          set -o pipefail
+          pixi update --json | pixi exec pixi-diff-to-markdown >> diff.md
+      - name: Create pull request
+        uses: peter-evans/create-pull-request@v7
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: Update pixi lock file
+          title: Update pixi lock file
```
mj-will: Can these be changed to match the commit message style we use? I guess this is either BLD or CI?

mj-will (changes requested):

.github/workflows/update-lock-files.yml
ColmTalbot: Done

mj-will (approved):

PR #1146, opened 2026-09-04
Add Welch PSD method option

JoeMetson:
`bilby_pipe` allows for the Welch PSD averaging method to be defined so this adds this option to `bilby` to keep consistent. Allows for the averaging method to be defined following the available methods in `gwpy`: welch, median or bartlett.

### DIFF
--- a/bilby/gw/detector/strain_data.py
+++ b/bilby/gw/detector/strain_data.py
@@ -365,7 +365,7 @@ class InterferometerStrainData(object):
 
     def create_power_spectral_density(
             self, fft_length, overlap=0, name='unknown', outdir=None,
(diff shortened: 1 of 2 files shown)


bilby/gw/detector/strain_data.py
```diff
@@ -414,8 +417,11 @@ def create_power_spectral_density(
         logger.info(
             "Tukey window PSD data with alpha={}, roll off={}".format(
                 psd_alpha, self.roll_off))
+        if method == 'bartlett' and overlap != 0:
```
Bill-M-Dev, line 420: Shouldn't this have a unit test, because it is a new piece of functionality?

PR #1149, opened 2026-09-18
Make sure the ROQ waveform generator wrapper is picklable

ColmTalbot:
@asb5468 pointed out that the ROQ likelihood is failing in py3.14.
This fixes a pickle issue when using forkserver with multiprocessing.

I'm not sure what a useful test looks like here.


mj-will (changes requested):
I think we should add a small test for this.

